PR #69, opened 2026-07-19
Expand effect-layer library: 10 new WGSL effects + visual preview grid

mysteropodes:
## Summary
- Adds 10 new effect types to the Calque d'effet system: sepia, invert, grayscale, posterize, pixelate, chromatic aberration, scanlines, film grain, sharpen, edge detection (Sobel) — all reimplemented from well-known public techniques in a single new shared WGSL pipeline (`simple_fx.wgsl` + `create_simple_fx_pipeline`/`simple_fx_pass` in `engine.rs`), rather than one dedicated pipeline per effect.
- Replaces the plain effect-type `<select>` with a 14-tile visual preview grid (CSS-filter approximations + titles) in the right panel; select stays in the DOM (hidden) as the single source of truth existing listeners already read/write.
- New effects share a generic 2-slot param-row UI (dynamic label/min/max/unit per type) instead of ~14 more dedicated param rows.

## Test plan
- [x] `cargo check --target wasm32-unknown-unknown`
- [x] `wasm-pack build --target web --out-dir ../src/wasm`
- [x] `node --check src/js/timeline.js`
- [x] In-browser: created an effect layer, clicked through tiles — chromatic aberration, edge detection, posterize, scanlines all produce correct live GPU results; edge-detect correctly outlines just the shape edges on black
- [x] Generic param rows show correct dynamic label/unit/value per selected effect type
- [x] Console clean throughout

## Deferred
The "fetch more effects from the web" button is out of scope for this PR — loading arbitrary third-party WGSL at runtime needs its own licensing/architecture review (explicit user decision this session).

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


mysteropodes:
Superseded by a rebased branch to resolve merge conflicts — see the new PR.

PR #203, opened 2026-07-25
Motion: split↔merge round-trip, undo inside components, parenting column + pickwhip

mysteropodes:
Reported: *"quand on double clic sur un calque avec plusieurs vecto/forme à l'intérieur impossible de revenir qu'à un seul calque après"* and *"tu n'as pas porté le parentage là-bas"*.

## The split had no inverse — and looking inside a component rewrote it

Double-clicking a Component layer in Motion auto-splits its contents into one layer per shape. That split existed in only one direction: **no merge command existed anywhere in the app**. Worse, the auto-split runs with `silent:true`, which also skips `pushUndo` — so merely *entering a component to look at it* permanently rewrote its layer structure with no undo entry.

- `mergeLayersIntoOne` is the true inverse. It concatenates the sources' strokes per frame with held-frame semantics (so merging layers whose keyframes don't line up doesn't silently drop content), demotes each source's layer-level Motion to per-element motion instead of dropping it, recovers the original name from `X — Forme N`, and re-points any layer parented to a disappearing layer at the survivor.
- Reachable from the layer context menu in **both** timelines, next to a new explicit "Éclater en calques", so the round-trip is discoverable from either end.
- The auto-split now records one undo step, and only when something will actually split.

## Undo inside a component was silently discarded

Found while testing the above, and much wider in scope than the split. `enterSymbol` aliases `state.layers` to `sym.layers`, so in-place edits write straight through — but `restoreLayersSnapshot` does `state.layers=[]` and rebuilds, severing the alias. **Every layer-level undo made inside a component** (split, merge, add/delete/reorder/rename a layer) was reverted the moment you returned to the scene. The screen showed the undone state right up until you left.

`exitToScene` now writes the array back as the authoritative chokepoint; `restoreLayersSnapshot` re-points the alias so other readers of the symbol agree in between.

## Parenting in Motion

Motion had parenting only as a dropdown buried in the properties panel, showing the active layer alone. Its layer rows now build the same `buildParentCell` Animation 2D uses. Both timelines additionally get AE's **pickwhip** — drag the dot onto another layer row, rubber-band line, highlighted drop target, Escape cancels, no undo entry on a dropped-on-nothing. It writes through `SMMotion.setLayerParent`, so the cycle refusal stays the single writer.

## Two dead gestures in Motion, fixed in passing

- Shift-click required a pre-existing selection while Motion's own plain click cleared it — so click / shift-click selected nothing, and multi-layer commands were reachable only via Cmd-click.
- The layer context menu opened only for a 2+ selection, so right-clicking a single layer did nothing at all.

## Verified by driving the app

Reproduced the original report (component → 4 layers, no way back), then round-tripped: merge restores one layer with all 4 strokes, the original name, the symbol storing one layer again, and the scene still rendering 4 strokes through the instance. Auto-split now undone by a single Cmd+Z, and the undo survives exit (`symAfterExit` matches the undone state). Pickwhip parents Layer 3 to Layer 1 and refuses the reverse (cycle); `parentChainMats` returns the live `(120,45)` offset for the child when the parent moves.


mysteropodes:
### Troisième lot : opérations de clés inspirées de Skew Pro

Ajouté après un tour d'horizon de [Skew Pro](https://www.goodboy.ninja/try/skew-pro/space) (sa liste de leçons est sa liste de fonctions). Nemo avait déjà les bords de skew, le déplacement en bloc, Distribuer, Flip, Inverser la sélection et « 1 sur 2 ». Manquaient :

- **Space** — la leçon exacte du lien : la boîte de sélection gagne ses bords **gauche/droite**, qu'on tire pour espacer ou resserrer les clés dans le temps, le bord opposé restant ancré. Branché sur le moteur de drag existant (le facteur devient par-CLÉ au lieu de par-ligne), donc il hérite de l'arithmétique absolue : zéro dérive, retour exact en revenant au point de départ.
- **Subdiviser** — insère une clé à mi-chemin de chaque paire, à la valeur que la courbe produit déjà. Vérifié non destructif : valeur identique à 3 décimales avant/après.
- **Sélectionner 1 sur N…** et **Garder au hasard…**

Deux défauts trouvés en testant ces ajouts, corrigés ici :

- Sur une sélection tenant sur **une seule piste**, la boîte fait ~11 px de haut : les bords de skew (14 px) la recouvraient entièrement et avalaient tous les clics destinés aux nouveaux bords latéraux. Les coins appartiennent maintenant aux bords latéraux.
- La boîte (fill + bords en `pointer-events:auto` pour être glissable) **avalait aussi le clic droit** sur les clés en dessous — or toutes les opérations par lot vivent dans ce menu contextuel. Dès que la boîte apparaissait, le menu devenait inatteignable pour la sélection même qu'il devait traiter. La boîte réémet maintenant l'événement vers ce qui est réellement sous le curseur.

Reste de la liste Skew Pro, non fait : Liquify (décalage pondéré par falloff), Align/Rotate sur la grille calques×temps.

PR #422, opened 2026-08-31
Canvas multi-select: boxes and hit-testing follow what is drawn

mysteropodes:
Chantier branch for the double-click / per-element selection work, from Cyril's brief: *"encore beaucoup d'erreur, entre les bounding box qui se mettent pas à jour en aller retour de group, si on ajoute des éléments dans le groupe entre paramétric shape ou brush, déplacement ciblage, zoom, position du canvas faut que ça soit au poil dans tous les cas de figure"*, plus *"bouger les layer à l'intérieur et voir les box… et la box globale se resize"*.

**Four real defects**, each measured before and after, in **both** modes.

### 1. Doubled boxes in Animation 2D
`motion.js`'s `perObjectBoxItems` had no mode gate, so it drew a second set of per-element boxes on top of engine-bridge's — two slightly different blues (`[74,158,255,190]` over `[63,107,245,120]`). At 1px it doesn't read as two boxes, just as a box that looks slightly wrong. My own regression from the #170/#176 work.

### 2. The group box ignored the layer's transform
`perObjectUnionBounds` — the region deciding "double-click inside to enter" / "click outside to leave" — was built from element-posed bounds only. Moving the **layer** left it behind while the drawn boxes followed:

| | boxes | enter/exit region |
|---|---|---|
| layer moved +400/−300 | moved | **unchanged** |

It now composes the layer transform once, at the end. `elementPosedBounds` deliberately stays element-only — both box builders apply the layer map themselves, so folding it in there would apply it twice.

### 3. Hit-testing tested geometry that had moved
A layer's Motion transform is applied at render and never touches the stored points; each element can carry its own on top. `layer.hitTest` cannot express that — it tests one point against geometry that several independent transforms have moved. `hitTestPosed` (tools.js) undoes the layer's transform once, then each child's own for that child alone.

Measured: layer moved/rotated/scaled **and** the two elements dragged apart → clicking either at its visible position selected **nothing**, in both modes. Wired into both double-click paths, select-bridge's Motion per-object block, and its main left/right click paths — the last of which also stopped gating the layer inverse on Motion mode, since `buildSceneJson` applies `motionMat` in either mode (CLAUDE.md §8).

### 4. The global box did not resize
Motion's `motionBoxGeom` built the layer box from `userLayers[i].bounds`, Animation 2D's `orientedSelBox` from raw path bounds. Moving elements apart left it at its old size while they spread outside it: drawn box unchanged at `[830,332,1772,447]` while the real extent had grown to `[510,-240,2112,1039]`. Both now build from posed bounds.

Animation 2D's is deliberately narrow — only when something in the selection actually carries an element transform — because `xformSelBounds` also feeds the panel fields and the move/scale **write** math, which work on raw geometry. No drag regression: deltas of 200/120 and −180/−100 land exactly, with and without a pose elsewhere.

## Verification
A matrix driven in **both** modes: 2 layers, each with a **parametric shape AND a brush stroke**; enter by double-clicking empty space inside the group, exit outside, re-enter; target each element; add an element while inside (box count and union grow, new shape targetable without re-entering); move elements apart individually then re-target them; move/rotate/scale the layer; **three zoom levels**; pan; then select an object on the **other** layer. All green.

Two of my own test artifacts, recorded so they are not re-found as "bugs": dispatching `mousedown` unconditionally after `pointerdown` replayed the Paper-native path and wiped the selection the pointerdown had just made (a real browser suppresses the compatibility mouse events when `pointerdown` is `preventDefault()`ed); and aiming at a wavy stroke's bounds **centre** misses the ink at high zoom, where the world-space tolerance shrinks.

**Still open on this branch:** text selection in a multi-layer canvas.

27/27 tests pass.

### DIFF
--- a/src/js/motion.js
+++ b/src/js/motion.js
@@ -5473,6 +5473,27 @@
     return null;
   }
   function onDown(event) {
+    // Inside a group with nothing picked yet, a click that lands ON a shape
+    // belongs to that shape (2026-08-31). Without this the whole-LAYER gizmo
+    // claimed it first — its ring, corners and body all sit over the very
+    // shapes you are trying to pick, and the more the layer is rotated or
+    // scaled the more of them fall under the cursor. Measured with the layer
+    // at rotation 25°: clicking a shape inside the group targeted nothing
+    // (_motionExpandedElement stayed empty) and the following drag moved the
+    // whole layer instead, which reads exactly as "I can select an object
+    // but not drag it".
+    //
+    // Deliberately narrow: only while per-object mode is on for THIS layer,
+    // only while no element is targeted yet, and only when the point is
+    // really on a shape. Every other Motion gesture — including the layer
+    // gizmo outside a group, and every grab once an element IS targeted —
+    // is untouched. Returning false hands the click to select-bridge, whose
+    // Motion block does the targeting.
+    if (window._perObjBoxes === state.activeLayerIdx && window._motionExpandedElement == null
+        && window.hitTestPosed && userLayers[state.activeLayerIdx]) {
+      var surForme = hitTestPosed(state.activeLayerIdx, event.point, 6 / Math.max(0.0001, view.zoom));
+      if (surForme && surForme.item && surForme.item.data && surForme.item.data.strokeId) return false;
+    }
     var ml = multiLayerBox();
     if (ml) {
       var mb = ml.bounds, mz = 1 / Math.max(0.0001, view.zoom);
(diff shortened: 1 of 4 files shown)


mysteropodes:
**Text in a multi-layer canvas — now verified too, no code change needed.**

Placed a real text object through the Text tool + `commitText` on the second layer, then drove both modes:

| | Animation 2D | Motion |
|---|---|---|
| counted as an element of the group | ✅ | ✅ |
| click selects/targets it | ✅ | ✅ |
| same with the layer moved + rotated | ✅ | ✅ |
| gets its own box in the group (3 boxes / 3 shapes) | ✅ | ✅ |
| then select a shape on the **other** layer | ✅ | ✅ |

One thing worth stating rather than "fixing": in Motion you must enter the group (double-click) before a single click targets an element — that is the mode's design, not a defect. My first Motion run failed only because it clicked before entering.

With this, every case from the brief is covered: round trip in/out of the group, adding an element while inside, per-element move then re-target, layer move/rotate/scale, three zoom levels, pan, the other layer, and text — in **both** modes, with parametric shapes **and** brush strokes.

PR #704, opened 2026-09-01
Replace the roadmap section with a real What's next

mysteropodes:
The old `The bigger goal` section named the eventual scope in one sentence and stopped. It described an ambition without giving anyone a reason to join it.

### What changed

Each remaining stage now says **what it is**, and where there is one, **the design decision behind it**:

- **Infinite canvas** — pre-production, where a project lives before it has a timeline
- **StoryBoard** — feeding straight into animation instead of export/re-import
- **Editing — node-based** — stated as a graph rather than a track timeline, *with the reasoning*: a shot in an animation pipeline is a graph of sources, versions and dependencies; stacked tracks force you to flatten that
- **Compositing — node-based** — same reasoning, one stage later
- **Sound design** — timed *with* the animation instead of against it

### The actual point of the change

The closing subsection. One person will not build an editor, a compositor and a sound stage — those are three separate crafts, not three more features. So it says that plainly rather than presenting a roadmap as a schedule, and tells someone whose craft that *is* that arriving early is when their input still shapes the design.

That's the difference between a roadmap that impresses and a roadmap that recruits.

### On the Blender comparison

Named as the reference for one app holding a whole pipeline. The claim is deliberately narrow — 2D has no **open-source** equivalent, which is defensible — rather than no equivalent at all, which isn't: proprietary integrated 2D suites do exist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,15 +88,57 @@ views of the same document.
   real production Kitsu instance — treat as experimental for now)*
 - Interactive in-app tutorial, 25 modules, in **English, French, Japanese, and Spanish**
 
-## The bigger goal
+## What's next
 
-Nemo aims to eventually cover the whole broadcast production chain in one application:
-pre-production, storyboard, editing, sound design, 2D animation, motion, and compositing.
+The goal is one application for the whole 2D production chain — so a project stops being a
+relay race between five programs that each own a piece of it and each disagree about what a
+frame, a layer or a timing actually is. Blender showed that a single app can hold a whole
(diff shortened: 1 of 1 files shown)


ivg-design (approved):
Reviewed: docs-only change to README.md. Merges cleanly onto current main (the `## The bigger goal` section is still there, unchanged since the branch point; later README edits touched only the header/badges/screenshot). The `#contributing` anchor resolves. Merging.

PR #710, opened 2026-09-01
Desktop feedback goes through the Worker, not a token in the binary

mysteropodes:
## ⚠️ Needs a desktop test before merging

`cargo check` passes and the JS parses, but **the transport itself is untested here** — it needs a running desktop build. This is also the exact area where a silent failure has already happened once (feedback saved locally, *appeared* sent, never left). Send one feedback from `npm run dev` and confirm it lands on `nemo` before trusting it.

---

The desktop build posted to GitHub straight from Rust with a fine-grained PAT **compiled into the binary** via `env!()`. That token shipped inside every copy of the app:

- it could be extracted
- it could not be rotated without publishing a new build
- pointing feedback at `nemo` would have meant shipping **`Contents: write` on the source tree** to every user

**Both previously issued tokens have since expired** — the same failure in slow motion. A binary already in testers' hands quietly stopped being able to report anything, and nothing could be done about it without a re-release.

### What changed

Both transports now use `workerPost()`. The only remaining token is the **Cloudflare secret inside the Worker** — never distributed, rotatable without touching the app.

### Why the Tauri HTTP plugin and not `fetch()`

This isn't a style choice. A plain `fetch()` would have been blocked **twice**:

- the app's CSP `connect-src` doesn't list the Worker
- the Worker's CORS allowlist holds the web origins, not `tauri://localhost`

Routing through the plugin issues the request from Rust, so neither applies — and **neither guard had to be widened**. Same pattern `kitsu.js` already uses.

### Verified

| Check | Result |
|---|---|
| `submit_feedback_issue` / `upload_feedback_attachment` removed | ✅ incl. `invoke_handler` |
| `STROKEMOTION_FEEDBACK_TOKEN` referenced anywhere | **0** |
| Rust compiles with *only* `STROKEMOTION_UPDATER_TOKEN` set | ✅ |
| JS syntax | ✅ |
| Runtime transport | ❌ **needs your desktop test** |

### DIFF
--- a/.github/workflows/deploy-feedback-worker.yml
+++ b/.github/workflows/deploy-feedback-worker.yml
@@ -10,10 +10,12 @@ name: Deploy feedback Worker to Cloudflare
 #   1. The Worker needs its OWN Cloudflare-side secret — NOT a GitHub
 #      Actions secret, set once via the CLI directly against the account:
 #        npx wrangler secret put GITHUB_FEEDBACK_TOKEN --name nemo-feedback
-#      (run locally, with a fine-grained GitHub PAT scoped ONLY to
-#      mysteropodes/strokemotion-feedback, Issues:write + Contents:write —
-#      the same scope src-tauri's STROKEMOTION_FEEDBACK_TOKEN already uses;
-#      can be the same token value or a fresh one).
+#      (run locally, with a fine-grained GitHub PAT covering both repos:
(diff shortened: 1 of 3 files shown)


mysteropodes:
Superseded by #874 — this branch had drifted 185 commits behind main and was unmergeable, so I redid the same idea fresh on top of current main instead of rebasing it.

PR #872, opened 2026-09-03
Fix release build: wire STROKEMOTION_FEEDBACK_TOKEN into release.yml

mysteropodes:
## Summary
- The v0.7.0-alpha.1 release build failed outright: `STROKEMOTION_FEEDBACK_TOKEN` is read via `env!()` at compile time (no runtime fallback), and `release.yml` never passed it in
- Wires the secret into the build step's env, alongside the Tauri signing keys
- Drops a stale comment block referencing PR #710's updater rewrite, which already landed

## Still needed before the next build
This repo has no `STROKEMOTION_FEEDBACK_TOKEN` secret yet — someone with repo admin access needs to:
1. Mint a fine-grained PAT at https://github.com/settings/tokens?type=beta scoped to **only** `mysteropodes/strokemotion-feedback`, with `Issues: write` + `Contents: write`
2. `gh secret set STROKEMOTION_FEEDBACK_TOKEN --repo mysteropodes/nemo`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #873, opened 2026-09-03
Rename STROKEMOTION_FEEDBACK_TOKEN to NEMO_FEEDBACK_TOKEN

mysteropodes:
## Summary
- Renames the compile-time env var/secret from `STROKEMOTION_FEEDBACK_TOKEN` to `NEMO_FEEDBACK_TOKEN` everywhere (lib.rs, docs, worker-feedback README, deploy workflow comment) — leftover naming from the StrokeMotion → Nemo rebrand
- Also renames the local dev-only `STROKEMOTION_FEEDBACK_ROOT` path override to `NEMO_FEEDBACK_ROOT`
- **Bigger fix found along the way**: desktop feedback (`submit_feedback_issue`) was still posting issues into the old `strokemotion-feedback` silo repo — the web path (`worker-feedback/src/index.js`) was already split (issues → `nemo`, attachments → `strokemotion-feedback`) and live, but the Rust side never got updated to match. Fixed to mirror the same split, with named constants.
- Docs (CLAUDE.md, deploy-feedback-worker.yml, worker-feedback/README.md) updated to describe the resulting token scope accurately: `NEMO_FEEDBACK_TOKEN` needs `Issues: write` + `Contents: write` on **both** `mysteropodes/nemo` and `mysteropodes/strokemotion-feedback` — which also means `Contents: write` on `nemo` itself, since a fine-grained PAT applies one permission set to every repo it covers. Flagged explicitly as a real scope increase from the original single-repo design.

## Depends on
#872 (adds the secret to release.yml) — merge that one first or alongside this one, both reference the same new name.

### DIFF
--- a/.github/workflows/deploy-feedback-worker.yml
+++ b/.github/workflows/deploy-feedback-worker.yml
@@ -1,19 +1,23 @@
 name: Deploy feedback Worker to Cloudflare
 
 # Separate from deploy-web.yml on purpose: this Worker holds a GitHub write
-# token scoped to mysteropodes/strokemotion-feedback (Issues + Contents),
-# the editor's static-site Worker holds none — no reason to redeploy one
-# whenever the other's files change.
+# token scoped to mysteropodes/nemo (Issues) + mysteropodes/strokemotion-
+# feedback (Contents, for attachments) — see index.js's ISSUE_REPO /
+# ATTACHMENT_REPO — the editor's static-site Worker holds none, so there's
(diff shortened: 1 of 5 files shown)


mysteropodes:
Superseded by #874 — #874 deletes the two Rust functions this PR was renaming the token for (submit_feedback_issue/upload_feedback_attachment), since desktop no longer embeds a feedback token at all.

Issue #902, opened 2026-09-04
[R06] Isolate task runtimes and mutable artifacts — tracking parent


ivg-design:
**Human accountable:** Cyrill (`mysteropodes`). **Role:** progress rollup; not an executable worker task.

[Approved English execution checklist](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.en.md) · [Français](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.fr.md). This replaces the former operative schedule and global blocking instructions. Prior comments, source/PR evidence and issue edit history remain historical evidence; the pre-reset body is also preserved in the local board-reset receipt.

Isolation source is integrated. Accept the bounded two-instance/native evidence separately; unrelated FFmpeg, signing or platform repairs do not block pure-unit extraction.

### Small executable outcomes

- [ ] [P14 — Complete two-instance native isolation acceptance](https://github.com/mysteropodes/nemo/issues/1016) — Cyrill/D2

### Closure and coordination

Only named leaf dependencies block work. Do not wait for all of R03/R05/R06 to finish before an independent extraction. Use native sub-issue progress and the leaf-only board views. Each leaf has one human assignee, lane, exact scope and observable completion checks; split work exceeding one bounded sprint before starting.

Baseline means exact observed state, including known defects. Source integrated, test evidence, installed/client acceptance and product bug repair are distinct. No new product feature, hosted Actions run or remote-agent infrastructure work is authorized by this parent.

This parent closes only after its agreed in-scope leaves are integrated and accepted. Deferred scope is not marked completed merely to remove it from the execution view. Current handoffs belong in the owning leaf; do not append another competing schedule here.

ivg-design:
The coordination-library increment and all five review defects are resolved in PR #935, merged as 4958912659d8bb0c1925b79333798e238d53e413. Candidate d520e949 passed 20 focused process/socket regressions plus doctor/check/verify (70 Node and 15 Rust tests). Keep #902 open: actual launcher/source-build handshake, concurrent browser/Tauri state, desktop/GPU reservation consumers, same-worktree build isolation and standard test discovery remain the next acceptance work. Creating isolated directories alone does not satisfy those application gates.

ivg-design:
[...]

PR #940, opened 2026-09-05
R06: isolate desktop build outputs and own their process lifetime

ivg-design:
Concurrent desktop builds in one worktree previously shared their Cargo output and could write generated source simultaneously. Add a standalone macOS build launcher that invokes the installed Tauri CLI with task-specific Cargo output, temporary paths and logs, and serializes builds by canonical worktree path. It provides source/build identity and owner-controlled status/stop while retaining completed artifacts until owner cleanup.

The build slot is released only after the complete owned process group is absent. Lead review reproduced a descendant surviving its build leader; the correction covers that sequence, including a descendant that ignores SIGTERM. Unconfirmed cleanup keeps the launcher and reservation live for reconciliation and owner retry. Other native platforms reject before creating task state.

Validation on `f1f378c8691db1a1520899a7d6db5a624944fc19`: six focused lifecycle tests passed; clean-commit doctor/check/verify passed with 185 Node and 15 Rust tests. Lead independently reran the original parent-exits-first reproduction and confirmed the descendant was gone. The installed macOS Tauri CLI help confirms the default build arguments. An 8 MiB-per-stream stub probe preserved both output files completely.

Advances #902. No actual native build was run in this increment. Paired native builds, generated-schema diff inspection, Tauri data/autosave, full cache-consumer coverage, and standard `build:desktop`/desktop-input/GPU integration remain open and coordinated with the existing R03/R04 owners. The wrapper sets `XDG_CACHE_HOME`; this is not evidence that every native cache consumer uses that directory.

### DIFF
--- a/engineering/runtime-isolation.md
+++ b/engineering/runtime-isolation.md
@@ -139,9 +139,13 @@ rules above. No unrelated browser or task is selected by name.
 Tauri CLI as `tauri build --target <host> -b app --no-sign`. It writes Cargo
 output under that task's `build/tauri-target`, routes temporary, cache and
 report paths through the task roots, and holds an exclusive slot derived from
-the exact worktree path while the build child is active. Different worktrees
-have different slots. A second build in the same worktree is refused before it
-starts, avoiding concurrent writes by Tauri/Cargo to shared generated source.
+the exact worktree path until the complete owned build process group exits.
+Different worktrees have different slots. A second build in the same worktree
+is refused before it starts, avoiding concurrent writes by Tauri/Cargo to shared
+generated source. Direct build-leader exit does not establish group completion:
+remaining descendants are terminated and verified absent before slot release.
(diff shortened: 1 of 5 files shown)


ivg-design (commented):
Reviewed exact head `f1f378c8691db1a1520899a7d6db5a624944fc19` and the clean-commit verification receipt. The process-group correction resolves the reproduced parent-exits-first leak: completion/release depends on the owned group being absent, and unconfirmed cleanup retains a live launcher and slot for owner retry. Output streams drain before the finite result is published.

Six focused lifecycle regressions and the recorded 185 Node / 15 Rust checks pass. Accept this bounded macOS launcher increment. Actual paired Tauri builds, native data/cache consumers, normal harness wiring, and desktop/GPU reservations remain open in #902; this does not establish full R06 acceptance.

Issue #901, opened 2026-09-04
[R05] Enforce module profiles and dependency boundaries — tracking parent


ivg-design:
**Human accountable:** Ilya (`ivg-design`). **Role:** progress rollup; not an executable worker task.

[Approved English execution checklist](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.en.md) · [Français](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.fr.md). This replaces the former operative schedule and global blocking instructions. Prior comments, source/PR evidence and issue edit history remain historical evidence; the pre-reset body is also preserved in the local board-reset receipt.

Retain integrated discovery, size profiles and no-growth ratchets. Full application/Rust dependency and global/cycle enforcement is not yet delivered; the leaves below separate these outcomes.

### Small executable outcomes

- [ ] [P11 — Cover Rust source classification and size ratchet](https://github.com/mysteropodes/nemo/issues/1013) — Cyrill/D1
- [ ] [P12 — Enforce the geometry engine public-module dependency boundary](https://github.com/mysteropodes/nemo/issues/1014) — Ilya/D2
- [ ] [P13 — Classify UI bootstrap/style/data separately from executable modules](https://github.com/mysteropodes/nemo/issues/1015) — Cyrill/D2
- [ ] [P33 — Accept the completed fixed census against integrated evidence](https://github.com/mysteropodes/nemo/issues/1035) — Ilya/O
- [ ] [B01 — Enforce the native application/MCP public-module boundary](https://github.com/mysteropodes/nemo/issues/1046) — Ilya/D2
- [ ] [T04 — Generate Rust coverage for the existing MCP crate](https://github.com/mysteropodes/nemo/issues/1053) — Cyrill/D1

### Closure and coordination

Only named leaf dependencies block work. Do not wait for all of R03/R05/R06 to finish before an independent extraction. Use native sub-issue progress and the leaf-only board views. Each leaf has one human assignee, lane, exact scope and observable completion checks; split work exceeding one bounded sprint before starting.

Baseline means exact observed state, including known defects. Source integrated, test evidence, installed/client acceptance and product bug repair are distinct. No new product feature, hosted Actions run or remote-agent infrastructure work is authorized by this parent.

[...]

PR #943, opened 2026-09-05
R05: add reviewed tooling boundary profile and baseline

ivg-design:
The boundary checker now has a reviewed profile and initial size baseline for 23 of the 24 files under `scripts/nemo/`. The profile records the real public entry points and permitted dependencies; a retained fixture makes all six violation categories reproducible.

Validation: the real profile and baseline pass with zero violations and three documented size warnings. The fixture exits 1 with seven violations covering cycles, private imports, forbidden layers, global state, size limits and expired exceptions. Scope, relative links, baseline equality and diff checks pass.

`receipt.cjs` remains explicitly excluded because its template expression triggers the current scanner’s ambiguous-slash refusal. This increment does not add application profiles or wire the profile into jobs/CI; #901 remains open for that remaining adoption work.

Related: #901.

### DIFF
--- a/engineering/boundaries/profiles/scripts-nemo.baseline.json
+++ b/engineering/boundaries/profiles/scripts-nemo.baseline.json
@@ -1,0 +1,71 @@
+{
+  "modules": [
+    { "id": "nemo.lib.util", "layer": "tooling-adapters", "dir": "scripts/nemo/lib",
+      "files": ["util.cjs"], "publicApi": ["util.cjs"], "sizeProfile": "Platform/engine adapter" },
+
+    { "id": "nemo.lib.identity", "layer": "tooling-adapters", "dir": "scripts/nemo/lib",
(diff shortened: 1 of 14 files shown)


ivg-design (commented):
Accepted this bounded tooling-profile increment at 025761b5bfe7af044b5008935e0f94d96833bef3.

The corrected layer assignments, public test entry points, actual dependencies, receipt exclusion and state-path ownership were checked against source. The two remaining prose errors found in final review were corrected in 025761b.

Independent execution: real profile plus baseline passes (23 modules, zero violations, three documented size warnings); retained fixture exits 1 with seven violations covering all six named rule categories. Baseline equality, relative links, scoped diff and private-marker checks pass. No checker or application behavior changed.

This accepts the increment only. The receipt scanner limitation, broader profiles and job/CI adoption remain tracked under #901. Integration uses the existing administrator exemption; the GitHub account cannot formally approve its own PR.

Issue #903, opened 2026-09-04
[R07] Install real PR and CI gates — tracking parent


ivg-design:
**Human accountable:** Ilya (`ivg-design`). **Role:** progress rollup; not an executable worker task.

[Approved English execution checklist](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.en.md) · [Français](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.fr.md). This replaces the former operative schedule and global blocking instructions. Prior comments, source/PR evidence and issue edit history remain historical evidence; the pre-reset body is also preserved in the local board-reset receipt.

Validation runs locally. No hosted Actions enablement, dispatch or rerun is authorized by ordinary commits/PRs. Adopt the real named checker/report jobs incrementally.

### Small executable outcomes

- [ ] [P09 — Wire capability/schema drift checks into normal local validation](https://github.com/mysteropodes/nemo/issues/1011) — Cyrill/D2
- [ ] [P10 — Enforce application dependency edges through the normal gate](https://github.com/mysteropodes/nemo/issues/1012) — Ilya/D2

### Closure and coordination

Only named leaf dependencies block work. Do not wait for all of R03/R05/R06 to finish before an independent extraction. Use native sub-issue progress and the leaf-only board views. Each leaf has one human assignee, lane, exact scope and observable completion checks; split work exceeding one bounded sprint before starting.

Baseline means exact observed state, including known defects. Source integrated, test evidence, installed/client acceptance and product bug repair are distinct. No new product feature, hosted Actions run or remote-agent infrastructure work is authorized by this parent.

This parent closes only after its agreed in-scope leaves are integrated and accepted. Deferred scope is not marked completed merely to remove it from the execution view. Current handoffs belong in the owning leaf; do not append another competing schedule here.

ivg-design:
## R07 tracking delta — PR #945 prerequisite

PR #945 is open at `f728346a5592d37edd5ba7ce4891afe7d9053e23`. It fixes the six runtime fixture failures caused by spaces in Buzz's Node path, with focused wrappers passing 14/14 and required quick verification passing 185 Node tests plus 15 Rust tests.

This is a test-only prerequisite increment and does not close #903. The issue remains open and blocked while the full CI integration and the unresolved prerequisite acceptance/receipt are reconciled.

PR: https://github.com/mysteropodes/nemo/pull/945

[...]

PR #934, opened 2026-09-05
Rebuild the LGPL ffmpeg sidecar against current Homebrew; build:desktop --no-sign (R04)

ivg-design:
Closes #900. Parent #888 (F0). Depends on #932 (merged).

## What changes

- `src-tauri/binaries/ffmpeg-aarch64-apple-darwin` rebuilt with the existing `scripts/rebuild-ffmpeg-lgpl.sh`, unchanged (FFmpeg n8.1, no `--enable-gpl`, `--enable-videotoolbox`). Codec, dependency and distribution policy are untouched: the linked library set is the same 26 Homebrew dylibs; only two sonames move (`libvpx.11→12`, `libSvtAv1Enc.3→4`) and the system `Metal.framework` is now referenced. `THIRD_PARTY_NOTICES.md` needs no change.
- `scripts/nemo/lib/jobs.cjs`: `build:desktop` passes `--no-sign`. Without it tauri builds `Nemo.app` and then fails signing the updater artifact for lack of the personal `TAURI_SIGNING_PRIVATE_KEY`, which a local verification build never has.
- `CLAUDE.md` §7 step 3: the sidecar is bound to the Homebrew *major* versions it was linked against; how to diagnose (`npm run doctor`) and repair (rerun the rebuild script).

## Reproduction (before repair)

Recorded on #900. On base `f529907` the sidecar failed on `libtheoraenc.2.dylib` (theora 1.1.1 installed); after `brew upgrade theora` (1.2.0, which also bumped libvpx and svt-av1) on base `e1d2ea7` it failed on `libvpx.11.dylib` and `libSvtAv1Enc.3.dylib`, and `bundle-ffmpeg-dylibs.py` exited with "dependency not resolvable".

## Validation (macOS 15 / Darwin 25.6.0 arm64, M3 Ultra, node 24.18, cargo 1.91.1, Homebrew theora 1.2.0 / libvpx 1.17.0 / svt-av1 4.2.0)

| Check | Result |
|---|---|
| Rebuilt sidecar `-version` | runs; `License: LGPL version 2.1 or later`; `--enable-gpl` absent from configuration |
| `otool -L` external dylibs | 26/26 resolve on this machine (`npm run doctor`: 0 missing) |
| Encoders present | `h264_videotoolbox`, `hevc_videotoolbox`, `prores_videotoolbox`, `prores_ks`, `libvpx`/`libvpx-vp9`, `libaom-av1`, `libsvtav1`, `libtheora` |
[...]

Issue #895, opened 2026-09-04
[R00] Resolve kickoff baseline and active work


ivg-design:
Kickoff has one current base, reconciled active work, explicit supported-platform candidates, and no conflicting writer scopes.

## Scope

Refresh main and open work; collect branches, heads, dirty/unpushed state, paths, contracts, sessions, and next checkpoints; identify the integration-owner and backup decisions still required.

## Acceptance

- [x] Current main, active PRs, local work, and collision candidates are recorded from a fresh read.
- [x] Every proposed writer scope has a base, branch/worktree, exact paths, dependencies, and next checkpoint.
- [x] No assignment, platform commitment, or authority is inferred where the maintainer has not supplied it.

## Dependencies

None. Kickoff ownership and scope decision is recorded in #895.

Parent: [F0](https://github.com/mysteropodes/nemo/issues/888)

## Execution checkpoint — 2026-09-05

- Execution: Clauditron recorded the kickoff receipt; Ilya accepted solo weekend kickoff.
- Current gate: Closed acceptance is unchanged; collaborator reports were waived, not received.
- Validation: **Accepted**. Original calendar dates remain forecasts; overdue work has no revised completion commitment.

- [x] Reconciled baseline and scopes; owner decision and admin integration route recorded.
- [x] Acceptance/waiver: https://github.com/mysteropodes/nemo/issues/895#issuecomment-5551299507.

## Planning and ownership

- Status: **Done** — current execution checkpoint below.
- Accountable human / integration owner: Ilya (`ivg-design`), under the accepted solo-weekend kickoff decision in #895.
- Execution coordination / review: Codexitron; existing Clauditron and delegated task ownership is preserved below.
- Program: Core; phase: F0; priority: P0; size: S.
- Initial proposal window: H0-H4; reviewed calendar forecast below. Effort estimate remains uncommitted.
- Before Ready: name the owner/reviewer, inspect current main and overlapping work, confirm bounded paths and required predecessor receipts.
- Before work: record current base, branch/worktree and agent/session in a scoped task receipt. Repository access follows the active dedicated Nemo workspace contract.
- Before Done: attach exact candidate/artifact, fixtures, platform results, remaining limitations and the applicable maintainer acceptance.

## Calendar forecast

- Schedule basis: **Forecast**, anchored to the user-selected September 4, 2026 start; planning reference **20:00 EDT (America/New_York, UTC-04:00)**.
- Start: **2026-09-04T20:00:00-04:00** (H0).
[...]

PR #951, opened 2026-09-05
test: isolate browser runtime source mutations

ivg-design:
Concurrent browser and build wrapper tests share the source checkout. The browser suite temporarily adds a root marker and an escaping symlink; either can invalidate a build launcher’s source handshake. A synchronized reproduction against `b698291` returned status 1 with `source moved: recorded source identity differs from current checkout`, while build identity still matched and the build was completed.

Run browser CLI fixtures from a private Git source root containing the actual runtime code and served assets. Extend the source-change test to keep a real build launcher in the original checkout alive and verify its owner/source/build handshake while the browser marker is present. Include the sanitized handshake reason in the build owner-status assertion. Production source and ownership checks remain enforced.

Validation:
- Concurrent runtime wrappers: 14/14 passing in two runs.
- Regression control: restoring the shared source root makes the new assertion fail with the reproduced source-moved reason.
- `npm run verify -- --profile quick`: 185 Node tests and 15 Rust tests passed; all four jobs passed.
- Module boundaries: 0 violations (3 existing size warnings); `git diff --check` passed.

Related: #903, #902, #947. This fixes the shared-checkout test race observed in PR #947; it does not close either remediation acceptance issue.

### DIFF
--- a/scripts/nemo/browser-runtime.test.cjs
+++ b/scripts/nemo/browser-runtime.test.cjs
@@ -5,16 +5,46 @@ const fs = require('node:fs');
 const http = require('node:http');
 const os = require('node:os');
 const path = require('node:path');
-const { spawn } = require('node:child_process');
+const { spawn, spawnSync } = require('node:child_process');
 const { after, test } = require('node:test');
 
 const scratch = fs.mkdtempSync(path.join(os.tmpdir(), 'nemo-browser-preview-test-'));
+after(() => fs.rmSync(scratch, { recursive: true, force: true, maxRetries: 3 }));
 process.env.NEMO_ISOLATION_ROOT = path.join(scratch, 'runtime');
+const sourceRoot = path.resolve(__dirname, '..', '..');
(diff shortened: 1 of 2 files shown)


ivg-design:
Accepted after independent source and negative-control review at `93266464ca79f48c2feb211e027d1af26d2f08f2`. Browser marker/symlink changes stay in a private Git source fixture; the new regression runs the real peer build CLI and preserves actual owner/source/build checks. Cleanup attempts both owner-controlled shutdowns and removes the private scratch root. Existing wrong-owner, stale-process and traversal assertions remain.

The synchronized pre-fix control fails correct-owner status with `source moved` while build identity matches and runtime is completed; restoring the shared root also makes the new regression fail. Corrected wrappers pass 14/14 twice; required quick verification passes 185 Node and 15 Rust tests. No production handshake was weakened. This resolves the specific hosted #947 race, not full R06 application/native acceptance.

Dry-run merge against main after #934 is clean. Integration follows the already accepted solo-weekend maintainer procedure in #895; this records agent review rather than a separate-account GitHub approval.

PR #952, opened 2026-09-05
R06: isolate native app data and WebKit website data per task

ivg-design:
Closes the last native item [`engineering/runtime-isolation.md`](../blob/main/engineering/runtime-isolation.md) left open for R06 ([#902](https://github.com/mysteropodes/nemo/issues/902)):

> Configure and exercise isolated Tauri data/autosave paths. Merely creating a `tauri-data` directory does not make the app use it. The platform owner must implement and validate the actual runtime override.

`lib/isolation.cjs` has created that `tauri-data` directory since the first R06 increment and nothing pointed the app at it. Two halves close it, sharing one environment contract and nothing else: `src-tauri/src/task_runtime.rs` (app side) and `scripts/nemo/native-runtime.cjs` (launcher side). Full design notes in `engineering/runtime/native-isolation.md`.

## Why the WebKit half is not optional

Redirecting `appDataDir()` alone would look like isolation and not be it. Autosave (`nemo-auto`), the recents list, the sync-folder setting and the feedback fallback store all live in `localStorage` — WKWebView website data keyed by the **bundle** identifier, which `appDataDir()` does not control. Two instances redirected only on the filesystem would still share one `localStorage`, so instance B's autosave would overwrite instance A's. `WebviewWindowBuilder::data_store_identifier` gives each window its own `WKWebsiteDataStore`.

That has to happen while the window is built, and Tauri builds `tauri.conf.json`'s windows *before* the setup hook runs. On the isolated path only, `run()` clears each window config's `create` flag (the one field Tauri's creation loop filters on) and `install()` rebuilds them with the identity and the same labels. **On the default path nothing is touched** and Tauri creates the window exactly as it does today.

## Two decisions worth reviewing

**`NEMO_TASK_DATA_ROOT` is the sole activation trigger.** `build-runtime.cjs` already exports `NEMO_TASK_ID` into the environment of every process it spawns, so keying off the task id could flip an unrelated launch into isolation by inheritance. Covered by `a_task_id_alone_never_switches_a_normal_launch_into_isolation`.

**A malformed isolation request exits 78 (`EX_CONFIG`); it does not fall back.** Silently using the shared install when isolation was requested writes a task instance's history, autosave and preferences into the real user profile — the exact failure the feature exists to prevent.

Also worth a look: the static fs capability scopes `$APPDATA`/`$TEMP`/`$HOME`, and an isolated root can sit outside all three, so `install()` widens the scope at runtime via `FsExt::fs_scope().allow_directory`. Without it the frontend's own `fs` calls are denied by the plugin with nothing visible in the UI.

`project.js` and `feedback-bridge.js` both resolve their base through one function (`SMProject.appDataBase`). Two readers of a new path is CLAUDE.md §1's first bug family; version history following the task root while feedback stayed in the shared install would be that bug.

## Validation

| Check | Result |
|---|---|
| `npm run verify` @ `38fd3f7`, clean tree | **PASS** — doctor PASS, check 6/6, `test:unit` 196/196, `test:rust` PASS. Receipt `20260905T215717Z-38fd3f7`. |
| `cd src-tauri && cargo test --lib task_runtime` | **15/15** |
| `node --test scripts/nemo/native-runtime.test.cjs` | **11/11** |
| `cargo check` (src-tauri) | clean, no new warnings |

The 15 Rust tests cover: the default path staying default, a task id alone never activating isolation, two instances getting non-overlapping roots, roots hanging off the declared base, and refusal-rather-than-normalization for every malformed or hostile request (relative path, `..`, `.`, filesystem root, one-level-deep root, empty, blank, NUL byte, bad task ids, nil/short/non-hex UUID), plus owner-only release, a refused release leaving state intact, a tokenless instance refusing every release, and re-validation of the root immediately before deletion.

The 11 Node tests cover: disjoint roots and distinct website-data identities, the emitted environment satisfying the app-side rules, valid non-nil v4 UUIDs, `.app` bundle resolution, two live concurrent instances each writing their record only into their own root, independent stop authority, owner-refused status/stop leaving the app and its state untouched, an unconfirmed-isolation handshake, and the exclusive `desktop-input` slot.

One of them found a real bug in my own validator: `Path::components()` silently drops `.` segments, so the `.`-rejection never fired and `/tmp/./x` arrived pre-normalized. Now scanned on the raw string.

## Remaining gate — paired two-instance desktop run

**Explicitly open, not claimed.** Everything above is pure-logic and launcher-level. That a *real* Nemo resolves these roots, and that two WKWebView website-data stores are genuinely separate on disk, needs a built app, two launched instances and the exclusive desktop input slot — which R04 currently holds (confirmed with Codexitron; waiting on its explicit handoff). Browser evidence cannot substitute for it.

Two smaller limits, reported rather than inferred:
- `data_store_identifier` needs **macOS ≥ 14** and Tauri returns no result for it, so `webDataStore.applied` means "the webview was built with this identity", not "WebKit accepted it". `observedPaths` is the only app-side evidence and is reported as observed-or-not.
- `declaredSource` is echoed verbatim from the launcher: it identifies the launcher that configured the instance, not the checkout the binary was compiled from. The app's own build identity is reported separately.

## Wiring requested from other owners (not applied here)

- **`package.json` / `lib/jobs.cjs`** (Codexitron / integration): nothing points at `scripts/nemo/native-runtime.cjs` yet. Also, `test:rust` runs `geometry-wasm` only — the 15 `src-tauri` tests in this PR are not reached by `npm run verify`.
- **`engineering/boundaries/profiles/scripts-nemo.profile.json`** (Claudimator Beta): no module entry for `native-runtime.cjs` or its test, so the boundaries checker does not see them.

## Two pre-existing issues found in passing, outside these paths

1. `scripts/nemo/build-runtime.test.cjs`'s "leader exit cannot release the slot while its stubborn descendant remains" is **flaky**: 6/6, 6/6, then 4/6 on three consecutive runs here. It waits at most 1s (100 × 10 ms) for the descendant's pid file, and on a loaded machine the IPC handshake can exceed that. Not touched by this PR — it uses none of these files.
2. `src-tauri/gen/schemas/capabilities.json` is committed **stale**: it lacks `dialog:allow-ask` and `dialog:allow-confirm`, which are present in `src-tauri/capabilities/default.json`. Any `cargo` invocation regenerates it and dirties the tree. Reverted here to keep this PR scoped; `src-tauri/gen` is not one of these paths.

Refs #902

### DIFF
--- a/engineering/runtime/native-isolation.md
+++ b/engineering/runtime/native-isolation.md
@@ -1,0 +1,194 @@
+# Native app-data and WebKit isolation (R06)
+
+[R06 / #902](https://github.com/mysteropodes/nemo/issues/902). This closes the
+one item [`engineering/runtime-isolation.md`](../runtime-isolation.md) left open
+for the desktop app:
+
+> Configure and exercise isolated Tauri data/autosave paths. Merely creating a
+> `tauri-data` directory does not make the app use it. The platform owner must
+> implement and validate the actual runtime override.
+
+`scripts/nemo/lib/isolation.cjs` has created a per-task `tauri-data` directory
+since the first R06 increment, and nothing pointed the app at it. Two pieces
+close that gap and share exactly one thing — an environment contract:
+
+| Half | File | Role |
+|---|---|---|
+| App | `src-tauri/src/task_runtime.rs` | Resolves the isolated roots, applies the WebKit website-data identity, widens the fs scope, writes an instance record, and serves `nemo_task_runtime` / `nemo_task_runtime_release`. |
+| Launcher | `scripts/nemo/native-runtime.cjs` | Allocates a task's roots and website-data UUID, spawns the app with that environment, and owns status/stop/cleanup plus the exclusive desktop-input slot. |
+
+## Why the WebKit half is not optional
+
+Redirecting `appDataDir()` alone would look like isolation and not be it.
+`nemo-auto` (autosave), `nemo-recents`, the sync-folder setting and the
+feedback fallback store all live in `localStorage`. On macOS that is WKWebView
+website data, keyed by the **bundle** identifier — nothing `appDataDir()`
+controls. Two instances redirected only on the filesystem would still share one
+`localStorage`, so instance B's autosave would overwrite instance A's.
+
+`WebviewWindowBuilder::data_store_identifier([u8; 16])` gives the window its own
+`WKWebsiteDataStore`. It needs the window to be built in Rust, and Tauri creates
+`tauri.conf.json`'s windows itself *before* the setup hook runs. So on the
+isolated path only, `run()` clears each window config's `create` flag — the one
+field Tauri's own creation loop filters on — and hands the configs to
+`install()`, which rebuilds them with the identity and the same labels. The
+entries stay in the config, so anything else reading `app.windows` still sees
+them. On the default path nothing is touched and Tauri creates the window
+exactly as it does today.
+
+## The environment contract
+
+| Variable | Required | Meaning |
+|---|---|---|
+| `NEMO_TASK_DATA_ROOT` | activation trigger | Absolute path, no `.`/`..` segment, at least two directories deep. Absent ⇒ the app resolves its normal install paths. |
+| `NEMO_TASK_ID` | when isolated | 1–120 ASCII letters, digits, `.`, `_`, `-`, starting alphanumeric — the same grammar as `isolation.cjs`'s `validId`. |
+| `NEMO_TASK_WEB_DATA_UUID` | when isolated, on macOS | 16-byte hex UUID, dashed or not, never nil. |
+| `NEMO_TASK_OWNER_TOKEN` | optional | Required to later call `nemo_task_runtime_release`. |
+| `NEMO_TASK_SOURCE_IDENTITY` | optional | R02 `sourceIdentity()` JSON, echoed back verbatim by the app. |
+
+**The data root is the sole trigger, deliberately.** `build-runtime.cjs`
+already exports `NEMO_TASK_ID` into every process it spawns, so keying off the
+task id could flip an unrelated launch into isolation by inheritance. A
+regression covers exactly that (`a_task_id_alone_never_switches_a_normal_launch_into_isolation`).
+
+**A malformed isolation request exits; it does not fall back.** `run()` prints
+the named reason and exits 78 (`EX_CONFIG`). Falling back to the shared install
+would write a task instance's history, autosave and preferences into the real
+user profile — the exact failure the feature exists to prevent.
+
+## Roots
+
+```
+<NEMO_TASK_DATA_ROOT>/
+  instance.json     written by the app: task id, pid, version, identifier, executable, uuid
+  data/             replaces appDataDir() for every JS consumer
+    history/        version snapshots (src/js/project.js)
+    autosave/
+    preferences/
+  cache/
+  output/
+  webkit/           records this instance's WebKit identity (see below)
+```
+
+`data/` keeps the layout `project.js` already writes, so `historyDir()` stays
+`<base>/history/<projectKey>` and only the base moves. The launcher points
+`NEMO_TASK_DATA_ROOT` at `taskRoots(taskId).tauriDataDir`, the directory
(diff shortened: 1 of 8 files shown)


ivg-design:
Paired native acceptance complete; PR head now **e798788** (rebased onto R04-integrated main, `3a3038d`).

## Candidate

`npm run build:desktop` PASS from this branch — receipt `reports/20260905T220451Z-30d3492`, `Nemo.app` at `src-tauri/target/aarch64-apple-darwin/release/bundle/macos`, bundle `Nemo` sha256 `e5447b39565a5ca6850fb99eca3a8b6d36c932a8209fd4bca8de8f8991d0cf43`, bundled `ffmpeg` sha256 `91566ff440e785f3f8e7799eb2d5f346c8b3e62ba8c3290f57d8edbf1fbb98e1`, committed sidecar `8fbf0afc…`. macOS 15 / Darwin 25.6.0, `desktop-input` slot held for the run and released after it.

## Two instances, real input, real state

`r06-paired-a` (pid 90343) and `r06-paired-b` (pid 90640) ran concurrently, each writing its own `instance.json` naming its task, pid, `com.strokemotion.app` and its own website-data UUID (`39590131-…` / `ef9a6ee4-…`). Driven only by System Events keyboard/pointer input:

- **Distinct saved state** — A set to 47 frames / 12 fps, B to 83 frames; the 30 s autosave landed as version history under each task's own `data/history/untitled-autosave/` (A `fps 12, totalFrames 47`; B `fps 24, totalFrames 83`).
- **WebKit on disk** — two stores under `~/Library/WebKit/com.strokemotion.app/WebsiteDataStore/<uuid>/`, one per instance. Neither instance's first launch offered a Resume card, i.e. each started on an empty `localStorage`.
- **Reload** — A quit with ⌘Q and relaunched on the same task id: same root, same UUID, start screen now offering *Resume Last Session*, restoring 47 frames / 12 fps.
- **Independent cleanup** — `stop` on A left 0 files under its root and removed exactly its own store; B kept running with its state, and B — relaunched *after* A's store was deleted — still resumed its own 83 frames. `stop` on B removed B's store; no store from this run remains.
- **Wrong-owner shutdown** — `stop --task r06-paired-b --owner <A's token>` refused (`owner token mismatch`, exit 1) and `status` reported `ok:false`, with both apps still running.
- **Shared install** — `~/Library/Application Support/com.strokemotion.app` byte-for-byte identical before and after.

## Two defects found and fixed

- **49085bf** — `launchConfig()` minted a fresh website-data UUID per launch, so a relaunched task got an empty `localStorage` (autosave, recents, sync folder) while its on-disk history was untouched, and every relaunch orphaned a store. The identity is now allocated once into `<task-root>/tauri-data/webkit/identity.json` (released with the roots), and `stop` removes the OS-side store after the app exits.
- **e798788** — `resolveExecutable()` took the first executable in `Contents/MacOS`, which is the **ffmpeg sidecar**; the first paired launch ran ffmpeg (exit 1 on its usage banner) while every launcher record read as a started isolated instance. `CFBundleExecutable` now selects the binary; a bundle declaring none is refused, not guessed.

Validation on the head: 15 cargo `task_runtime` tests, 14 node native-runtime tests (through normal discovery), `npm run verify --profile quick` PASS 4/4 (receipt `reports/20260905T223416Z-e798788`).

Note for the lead: #953 implements the same R06 scope with a different mechanism (suffixed bundle identifier vs. data-root env + `WKWebsiteDataStore` identity). These two should not both merge.

ivg-design:
Personal triage: keep open as the unselected native-isolation alternative; do not merge alongside #953/#963. This branch uses a different activation/environment and frontend path-resolution design. The existing integration lane explicitly chose #953 plus #958 and added further cleanup, handshake and acceptance corrections. No decision to delete or discard this alternative's history is made here. Final reconciliation/closure belongs with the selected #963 integration and its native owner.

ivg-design:

Not selected: this was the alternative native-isolation design (different activation/environment and frontend path resolution). The integration lane chose #953 + #958 (via #963), integrated through merged PRs #988/#991, and P14/#1016 two-instance native isolation acceptance is accepted on that design. Closing the unselected alternative; branch retained for reference.

PR #953, opened 2026-09-05
R06: isolate native app state per task

ivg-design:
Closes half of the remaining R06 work in #902: the native side of task-runtime isolation.

`scripts/nemo/lib/isolation.cjs` already allocated a `tauri-data` root per task, and `engineering/runtime-isolation.md` named the gap in as many words — *"Merely creating a `tauri-data` directory does not make the app use it."* Nothing consumed the root, so two concurrent instances of the packaged app shared one identifier, one set of app directories and one WebKit store.

## Native side — `src-tauri/src/task_runtime.rs` + bootstrap hunks in `lib.rs`

Two mechanisms, because macOS keeps the two kinds of state in two different places:

| State | Mechanism |
|---|---|
| App data / config / cache / logs | every `app_*_dir()` resolves as `<platform dir>/<config identifier>` (tauri 2.11.3 `src/path/desktop.rs`), so the identifier becomes `com.strokemotion.app.nemo-task-<16 hex>` before the builder runs. The fs plugin scope, `appDataDir()` in JS, feedback and history folders all follow without knowing this exists. |
| WebKit `localStorage` / IndexedDB | not under those directories at all — the configured windows are created in `setup` with an explicit `data_store_identifier` (macOS ≥ 14), using the documented `WindowConfig.create = false` path. Labels are untouched, so `capabilities/default.json`'s `"windows": ["main"]` still applies. |

Activation is `NEMO_TAURI_DATA_DIR`. Absent, **not one resolved path changes**. Present with anything else missing or invalid, the app exits 2 naming the variable — a silent fallback to the shared identifier would write an isolated run's state into the user's real app data. `NEMO_TASK_ID` alone is deliberately not enough, since the build launcher already exports it for every desktop build.

The task key (`idKey`, SHA-256 of the exact task id) is supplied by the launcher instead of recomputed in Rust, so there is one derivation rather than two that must stay identical by hand.

## Launcher side — `scripts/nemo/lib/native-runtime.cjs`, `scripts/nemo/native.cjs`

Starts a built Nemo with that environment, owns its process group, and reports the manifest the instance writes for itself (`native-runtime.json`, schema `nemo.native-runtime/1`, also the app's first stdout line, never containing the owner token). `status` is `ok` only when the owner token, the complete R02 source/build identity, and a live manifest naming *this* task all agree. `stop` is owner-only and removes the task's roots including `tauri-data`.

`--reserve desktop-input,gpu-reference` gives the shared desktop resources an explicit exclusive reservation with a real consumer. Running two isolated instances at once is *not* what a slot restricts.

`nemo_task_runtime` returns the manifest to the page; `nemo_task_runtime_release` takes the owner token and exits — the stop path that stays available when process-group termination does not (an instance started through macOS `open` is reparented by launchd).

## Validation

- `cargo test task_runtime` — 10/10: activation, six fail-closed refusals, uppercase-key refusal, identifier and data-store derivation, task-id rule mirrored from the Node library.
- `node --test scripts/nemo/native-runtime.test.cjs` — 6/6: two concurrent instances on disjoint roots with per-task keys, handshake refusal for a silent or foreign manifest, owner-only stop with root removal, reservation refusal and release, missing-build blocker, and the environment contract asserted against `task_runtime.rs` so renaming a variable on one side only fails the suite.
- `npm test` — 191/191, reaching the new suite through `tests/nemo-native-runtime.test.cjs` (existing `tests/*.test.cjs` discovery, no harness change).
- Real packaged release build via `node scripts/nemo/build.cjs start --task r06-native-build` (isolated `CARGO_TARGET_DIR`, `--no-sign`, `completed`, exit 0). That binary refuses all six invalid isolated environments with exit 2 **before any window is created**.

## Not covered here

Two live Nemo windows saving and reloading side by side. That needs the exclusive desktop input/GPU slot, currently held for R04 packaged UI export; the doc's remaining-integration list states this explicitly rather than implying the acceptance bullet is closed.

`src-tauri/gen/schemas/capabilities.json` is left dirty on purpose: it is build-regenerated drift (the committed copy is missing `dialog:allow-ask`/`dialog:allow-confirm` that `capabilities/default.json` already lists), pre-existing and shared with other worktrees.

Refs #902

### DIFF
--- a/engineering/runtime-isolation.md
+++ b/engineering/runtime-isolation.md
@@ -247,8 +247,20 @@ node scripts/nemo/native.cjs stop --task app-a --owner <token>
 build identities, and a live app manifest naming this task agree. A manifest
 left by an earlier or different run is reported as such, never accepted as this
 instance's identity. `stop` refuses a mismatched owner, terminates the owned
-process group, waits for launcher exit and then removes that task's roots,
-including its `tauri-data`. Copy wanted artifacts first.
+process group, waits for launcher exit and then removes that task's roots.
+Copy wanted artifacts first.
+
+Removing the task root is not enough on its own: the isolated app directories
+are `<platform dir>/<identifier>`, so they live outside it and
+`isolation.releaseTask` never sees them, as does the WebKit store under
+`~/Library/WebKit/<bundle identifier>/WebsiteDataStore/<uuid>`. `stop` therefore
+reads the manifest the instance wrote, and removes each directory **whose last
+component carries this task's own identifier suffix** — recomputed locally from
+the task id, never trusted from the manifest. A manifest naming the shared
+`com.strokemotion.app` directory, another task's directory or a relative path is
+refused by name and reported in `appState.refused`. A start that fails does not
+release anything it did not acquire: task ids are reused across restarts, which
+is what lets an instance's state survive a stop.
 
 `--reserve` takes the shared resources this instance needs exclusively, before
 it starts: `desktop-input` for a run that drives the one keyboard and mouse,
@@ -263,8 +275,10 @@ use an executable stub that reproduces only the app's observable contract (the
 environment it is handed, the manifest it writes), so they cover two concurrent
 instances on disjoint roots, per-task keys, handshake refusal for a silent or
 foreign manifest, owner-only stop with root removal, explicit reservation
-refusal and release, and the missing-build blocker — without a desktop, a GPU or
-a window. The environment contract itself is asserted against
+refusal and release, the missing-build blocker, derivation-bound removal of the
+isolated app directories (refusing the production directory, another task's
+directory and a relative path), and a refused start leaving an earlier
+instance's state intact — without a desktop, a GPU or a window. The environment contract itself is asserted against
 `src-tauri/src/task_runtime.rs`, so renaming a variable on one side only fails
 the suite instead of silently returning the app to shared state. The native
 resolution rules (activation, fail-closed refusals, identifier and data-store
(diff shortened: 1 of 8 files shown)


ivg-design:
## Paired two-instance live validation — done (2026-09-05, after the R04 slot handoff)

Two packaged instances of this branch's build, up at the same time, driven through `scripts/nemo/native.cjs`.

| Check | Result |
|---|---|
| Two concurrent app processes | pids 87295 and 87578, both `active`, from the same `Nemo.app` |
| Identifier per instance (Tauri's own resolver, read back from the running app) | `com.strokemotion.app.nemo-task-99377ca9e44b34f0` vs `…-80a136b6d8e6214d` |
| App directories | distinct Application Support / Caches / Logs / config paths per instance |
| WebKit store on disk | two real stores: `~/Library/WebKit/com.strokemotion.app/WebsiteDataStore/99377ca9-e44b-34f0-1efc-4ff5e006057f` and `…/80a136b6-d8e6-214d-7c77-837848125c47`, each with its own LocalStorage/IndexedDB/Origins tree |
| Real writes | both instances autosaved project JSON into **their own** `history/untitled-autosave/`; 9 files against 24, **zero shared filenames** |
| Real UI driven | instance A: New Project card → dialog → name typed → Create pressed, through the accessibility tree |
| Shared production state | untouched: `com.strokemotion.app` still 25 files / 464K with newest mtime **predating** the run; production WebKit (22 files / 1884K) and Caches (6 files / 244K) unchanged |
| State survives app exit | A stopped, its 9 autosave files retained |

`data_store_identifier` is therefore confirmed as more than "built with this identity" — macOS 26.6 materialised both stores under the bundle container, keyed by the UUID form of each task key.

**Not closed:** restarting the same task id and reading the saved document back through the UI. The relaunch was refused because a parallel session took the `desktop-input` slot (holder `r06-paired-a`, pid 12439) — the reservation behaving exactly as designed. I stood down rather than contend for it.

## Two defects the run found, both fixed in `74b4bd0`

1. **Isolated app state was never released.** Those directories are `<platform dir>/<identifier>` — outside the task root, so `isolation.releaseTask` never saw them, and neither did the WebKit store. Two cleanly stopped instances leaked all of it. `stop` now removes each directory whose last component carries this task's own identifier suffix, recomputed locally from the task id and never trusted from the manifest; a manifest naming the shared production directory, another task's directory or a relative path is refused by name. Verified against the real leftovers: both isolated roots and both WebKit stores removed, `com.strokemotion.app` refused.

2. **A refused start destroyed the previous run's state.** `runNativeLauncher` called `isolation.releaseTask` on failure paths (copied from the build launcher). Task ids are reused across restarts — that is what makes state survive a stop — so a start that never took ownership was deleting the earlier instance's records. It cost exactly that live: the refused `desktop-input` reservation wiped instance A's manifest and status.

`npm test` 194/194, `cargo test task_runtime` 10/10.

ivg-design:
Personal triage: selected implementation is retained in #963; do not integrate this older head separately. Exact `dafbc5ff62` is an ancestor of #963 `6a976614`. The successor includes important later ownership/handshake/cleanup and packaged-harness corrections. Preserve the original paired-instance evidence, but final native Open/Resume/render acceptance and the exact combined hosted result remain with the existing integration owner.

ivg-design:

Superseded: head `dafbc5ff` is an ancestor of #963 `6a976614`, whose retained implementation was integrated via merged PRs #988/#991. The 2026-09-06 triage already directed not to integrate this older head separately.

PR #958, opened 2026-09-05
R06: reconcile a released task's dead exclusive slots (#902)

ivg-design:
Refs #902 (R06 task-runtime isolation).

## The residue

Exclusive slot records live in `<runtime>/slots/`, deliberately outside every task root, so `releaseTask`'s `rmSync` of the task tree could never reach them. Reproduced against `origin/main` (d499521) with a disposable `NEMO_ISOLATION_ROOT`: register a launcher, acquire a slot, SIGKILL the launcher, then call `releaseTask` with the correct owner token.

```
releaseTask -> {"released":true,"taskId":"repro-task-59925"}
task root removed: true
RESIDUE slot files after authorized release: 1
  record: { "slot": "repro-slot-59925", "taskId": "repro-task-59925",
            "pid": 59926, "ownerToken": "...", "acquiredAt": "..." }
holder pid alive: false
```

`acquireExclusiveSlot` already reclaims a dead holder, so this never wedged a slot — the record just outlived the task that owned it, naming a released task and a dead pid until some other task happened to want that exact slot.

## The correction

An owner-verified release now reconciles only what the record itself proves:

- it names **this** task (`rec.taskId === taskId`),
- it is complete (`slot` and `ownerToken` present, `pid` a positive safe integer),
- its declared slot name is the one its filename encodes — so the per-slot guard taken below is the right guard, and a crashed `writeRecord` `.tmp` leftover is rejected rather than reclaimed,
- its holder pid is verifiably not running.

Every removal happens inside the same `mutate('slot', ...)` guard that `acquireExclusiveSlot`/`releaseExclusiveSlot` already take, and **re-reads the record under that guard**, so a contender that reclaimed it between the scan and the removal is never overwritten.

This cannot grant more than the existing acquire path already grants: a dead holder's record is reclaimable by any task today, so deleting one is strictly narrower than reclaiming it.

Anything live, foreign, malformed, self-inconsistent or under an interrupted guard is reported in the new `slots` result and **left exactly as found**. A release with no launcher record to check against has no ownership proof and touches nothing beyond its own tree — which keeps the existing ownerless `isolation.releaseTask(taskId)` calls in `lib/build-runtime.cjs` fail-closed.

The helper is **not exported**: the authority is the caller's verified launcher token inside `releaseTask`, not a standalone verb. The module's public API is unchanged, so the `nemo.lib.isolation` boundary profile needs no edit.

## Result shape

```json
{ "released": true, "taskId": "cli-probe",
  "slots": { "scanned": 1, "reconciled": ["cli-slot"],
             "retained": [], "otherTasks": 0, "truncated": false } }
```

Bounded at 4096 scanned entries with a truthful `truncated` flag.

## Validation

- `node --test scripts/nemo/isolation.test.cjs` — **24/24 pass**, including the pre-existing concurrent stale-reclaim and interrupted-mutation guarantees, which are untouched.
- `node scripts/nemo/ci.cjs quick` on this exact commit (head `797e64b`, base `d499521`) — `ok: true`, `problems: []`; doctor / check / test:unit / test:rust all pass.
- `node scripts/nemo/ci.cjs boundaries` — exit 0, ratchet `ok: true`, no violations.
- Repro re-run against the patched library: `RESIDUE slot files after authorized release: 0`, `slots.reconciled: ["repro-slot-..."]`.
- CLI end-to-end: `isolation.cjs release` prints the `slots` block; a live launcher is still refused with `launcher still running; stop it before release`.

New tests: killed-child + rightful-owner reclaim with a foreign task's dead slot left intact; live, malformed and filename-mismatched records retained with reasons; non-owner release reconciling nothing; interrupted slot guard blocking reconciliation without blocking the task release.

## Known limitations (both outside this branch's owned paths)

- `engineering/runtime-isolation.md` line 22 says release "removes only that validated task's roots after exit". That is now incomplete — it also reconciles that same validated task's verifiably-dead slot records. Left for that file's owner.
- `scripts/nemo/isolation.test.cjs` is now 479 nonblank lines, crossing the advisory `warn` threshold of 450 for the "Test file" profile (hard maximum 600). Advisory only; the boundaries lane passes, and several modules already sit above their warn thresholds on `main`.

All probes used disposable runtime roots. No native candidate, `task_runtime.rs`, native-runtime, package/jobs, profile, user Library, real slot or desktop state was touched.

### DIFF
--- a/scripts/nemo/isolation.cjs
+++ b/scripts/nemo/isolation.cjs
@@ -37,7 +37,9 @@ function usage() {
     '  roots     --task ID                                       print the root paths for an existing/new task id',
     '  handshake --task ID --owner TOKEN [--check-source]      verify a launcher is alive and matches',
     '  stop      --task ID --owner TOKEN [--signal SIG] [--timeout-ms N] stop a task; refused unless --owner matches',
-    '  release   --task ID [--owner TOKEN]                       remove task roots only after exit; existing records require their owner',
+    '  release   --task ID [--owner TOKEN]                       remove task roots only after exit; existing records require their owner.',
+    '                                                            An owner-verified release also reports `slots`: this task\'s verifiably-dead',
+    '                                                            slot reservations reconciled, and every record left as found with the reason.',
     '  slot-acquire --slot NAME --task ID [--pid N]              acquire a named exclusive slot (desktop input, GPU bench, ...)',
     '  slot-release --slot NAME --owner TOKEN                    release a named exclusive slot',
   ].join('\n'));
--- a/scripts/nemo/isolation.test.cjs
+++ b/scripts/nemo/isolation.test.cjs
@@ -412,6 +412,115 @@ test('simultaneous launcher registrations have exactly one persistent owner', as
   }
 });
 
+// ---- dead-task slot residue (slot records live outside every task root) ----
+
+// SIGKILL and wait until the pid is really gone, so a record naming it is
+// unambiguously dead rather than merely unresponsive.
+async function reap(child) {
+  child.kill('SIGKILL');
+  await waitForExit(child, 5000);
+  const deadline = Date.now() + 2000;
+  while (iso.pidAlive(child.pid) && Date.now() < deadline) await new Promise((resolve) => setTimeout(resolve, 20));
+  assert.equal(iso.pidAlive(child.pid), false);
+}
+
+// Take the slot legitimately (acquire refuses a dead pid, by design), then kill
+// the holder: the residue an interrupted run actually leaves behind.
+async function slotHeldByDeadPid(slot, task) {
+  const child = spawnDummyProcess();
+  const held = iso.acquireExclusiveSlot(slot, task, { pid: child.pid });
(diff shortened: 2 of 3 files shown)


ivg-design:
Personal triage: this bounded dead-slot correction remains selected in #963, with its existing independent 24-test evidence. The successor also composes later native/build cleanup corrections. This older standalone branch currently has failed affected-surface/aggregate checks, so I am retaining the selected combined route rather than bypassing that technical gate. Native/source ownership is unchanged; final integration should reconcile and close this predecessor after the successor lands.

Execution-policy update: the user's explicit local-build-only instruction supersedes the hosted-run requirement above. Clear the same technical acceptance gates with exact-candidate local evidence. Historical hosted failures remain evidence of missing validation; no new GitHub Actions run is required or authorized. All workflows remain disabled.

ivg-design:

Superseded: the dead-slot reconciliation was selected into #963 and integrated via merged PRs #988/#991 with its 24-test evidence. Both files exist on main in their later integrated form.

PR #993, opened 2026-09-07
Preserve live renderer and Labs state when project import is rejected

ivg-design:
Malformed project imports now fail before resetting Labs or clearing the active renderer path store. This adopts the application integration owner’s reviewed `d8c316de` correction with cherry-pick provenance and extends the real import regression to assert that both live states survive rejection.

The new regression fails on main after #991 (5 pass, 1 fail) and passes with this correction (6 pass, no failures/skips). All six source checks, inventory freshness and the protected-base boundary gate pass on the corrected composition. Coverage and inventory pins are refreshed for the changed timeline bytes; existing size ceilings and baseline remain unchanged.

The specific source correction received Codexalog’s signed PASS. #902 remains open for its native workflow and performance acceptance; this fixes the rejected-import defect only. No native package, hosted workflow or release claim.

Related: #991, #902.


ivg-design (commented):
Correction accepted on exact head 521fe316ebe7bfcfcf0d64bfc2892b9d5ab71f62. Independently reproduced the renderer-state regression against the integrated R06 source at 223f98d: the existing invalid-import regression failed before the parse-order change and passed afterward. Both malformed JSON and structurally invalid layers must leave renderer and Labs state untouched.

The complete src, tests, and src-tauri trees match independently validated candidate 2ee5e2cc93382f23d9d5dafa89cc65e0f8cd676b; differences are source provenance metadata in three generated records. Validation: 8 focused project tests, 561 Node tests (one existing skip), 15 geometry Rust tests, 5 browser tests, six source checks, 902-row inventory and protected-base boundary gate passed. Existing profile ceilings and inventory rows are unchanged.

This clears the import-correction review gate. It does not close #902 or establish packaged desktop/native performance acceptance.

Issue #1000, opened 2026-09-07
Committed ffmpeg sidecar links against Homebrew majors absent from the validation workstation (blocks bundle-ffmpeg-dylibs.py and DMG packaging)


mysteropodes:
Reproduced independently on the validation workstation at `origin/main` = `66ece0641708122eb8447e85ad8dd7e3402aaf6c`. Surfaced by the R14 packaging lane, but it is **not** an R14 defect — R14's MCP sidecar has zero Homebrew dependencies and is unaffected.

## Symptom

`scripts/bundle-ffmpeg-dylibs.py` hard-exits during desktop packaging:

```
ERROR: dependency not resolvable on this machine: /opt/homebrew/opt/svt-av1/lib/libSvtAv1Enc.4.dylib
```

`npm run doctor` reports `sidecar dylibs 26 external, 2 missing on this machine` (`libvpx.12.dylib`, `libSvtAv1Enc.4.dylib`) and `sidecar runs MISSING`.

The committed sidecar cannot execute at all:

```
$ ./src-tauri/binaries/ffmpeg-aarch64-apple-darwin -version
dyld: Library not loaded: /opt/homebrew/opt/libvpx/lib/libvpx.12.dylib
```

Net effect: no shippable DMG can be produced on this workstation, and the committed sidecar is non-runnable there.

## Verified facts

- Committed artifact: `src-tauri/binaries/ffmpeg-aarch64-apple-darwin`, tracked, blob `97b0232cd393bff08435b2b21217dd7c3772f184`.
- `otool -L` on it requires `/opt/homebrew/opt/libvpx/lib/libvpx.12.dylib` and `/opt/homebrew/opt/svt-av1/lib/libSvtAv1Enc.4.dylib`.
- Installed on the workstation: libvpx **1.15.2** (ships `libvpx.11.dylib`) and svt-av1 **3.1.2** (ships `libSvtAv1Enc.3.dylib`). Neither major is present.
- Last change to the artifact: `3a3038d` — *"Rebuild the LGPL ffmpeg sidecar against current Homebrew; build:desktop --no-sign (R04)"*, PR #934, merged 2026-09-05.

## The part that needs an owner's judgement

[...]

PR #1001, opened 2026-09-07
R14 Lane H: bundled MCP executable in an installed desktop package, plus clean attach proof

mysteropodes:
R14 Lane H — evidence that the `nemo-mcp` executable ships inside an installed Nemo desktop package, and that a client can attach to it and select the intended running instance with nothing else on its resolution path.

**Evidence only.** No source, build script or configuration file is changed. Everything is under `engineering/remediation/receipts/R14-packaging/`.

Base `66ece0641708122eb8447e85ad8dd7e3402aaf6c`. Built and installed locally on darwin/arm64. **No GitHub Actions workflow was enabled, dispatched or rerun at any point.**

## What was done

Built the package locally, ran the CLAUDE.md §7 dylib step, installed the result the way a user does — mount the DMG, copy `Nemo.app` out, eject — into `~/Applications`, deliberately not `/Applications`, so Cyril's existing 2026-07-20 build was not overwritten. Then attached to the bundled executable from a scrubbed environment: `env -i`, a single empty directory as the whole `PATH`, session running from `/`, no registry override set.

**Bullet 4.** The sidecar ships at `Contents/MacOS/nemo-mcp` and carries one digest (`93f346fa…`) at every stage: staged by `build-mcp-sidecar.cjs`, copied into the app by Tauri's `externalBin`, inside the disk image, and installed. It links only `/usr/lib/libiconv.2.dylib` and `/usr/lib/libSystem.B.dylib`. Homebrew-linked dependency count per bundled executable: `nemo` 0, `nemo-mcp` 0, `ffmpeg` 26.

**Bullet 5.** It completes an MCP handshake (`initialize` → `nemo` 0.1.0, protocol `2025-06-18`) and a capability listing (`tools/list` → `nemo_discover`, `nemo_query`, `nemo_command`). With four instances running, each `nemo_query` returned the `instanceId` it asked for **and a `documentId` unique to that instance**, both matching what `nemo_discover` reported moments earlier — the client never supplies `documentId`, so this discriminates rather than echoes. A never-registered UUID was refused with a specific message. Instance→PID attribution was established independently of the protocol, by mapping each record's port to its listening process with `lsof`.

The empty-`PATH` result only shows no toolchain resolves *by name*, so the running server was also inspected with `lsof` mid-session: `cwd=/`, the only mapped executables are the installed binary and `/usr/lib/dyld`, and **0** handles reference any source checkout.

## Findings, all outside this lane's write scope

1. **The pinned sidecar SHA-256 is not reproducible.** The release binary embeds its own absolute build path, so identical source built in different checkouts gives different digests (`93f346fa…` here, `2adbbf90…` in another checkout). Stable *within* a checkout. Either drop the pinned hash from acceptance or add `trim-paths = "all"` to `nemo-mcp/Cargo.toml`.
2. **`scripts/bundle-ffmpeg-dylibs.py` fails on this machine — pre-existing, not MCP-related.** The committed ffmpeg sidecar wants `libvpx.12` and `libSvtAv1Enc.4`; Homebrew here has `libvpx.11` and `libSvtAv1Enc.3`. `npm run doctor` reports `26 external, 2 missing` and `sidecar runs MISSING`. Exactly the drift CLAUDE.md §7 predicts, dating from R04. It does not affect `nemo-mcp`, which needs no bundled dylibs. **The script was deliberately not softened** — its hard exit is what prevents shipping an app that crashes on Export, and §7 records that softening it already cost a release once. Remedy is `rebuild-ffmpeg-lgpl.sh` plus committing `src-tauri/binaries/ffmpeg-aarch64-apple-darwin`, outside these paths.
3. **Discovery records outlive their instance.** A controlled test quit an instance through the normal path and its record remained, so the leak is routine, not just on crashes. `nemo_discover` is *not* affected — it probes and lists only responders (0 instances returned while a stale record sat on disk; 4 while 4 were live). The gap is in the non-probing path: `nemo_query`/`nemo_command` return `Connection refused (os error 61)` for a stale id, where a never-registered id gets the actionable `Selected instance is absent; discover again after reconnect`.
4. **The three tools disagree on the instance field name.** `nemo_discover` emits `instanceId`, `nemo_query` requires `instance_id`, `nemo_command` accepts `instanceId`. Feeding discovery output straight into `nemo_query` fails on a missing field. `Query` in `server.rs` lacks the `rename_all = "camelCase"` that `ApplicationRequest` has.

## Not covered

Updater signing (needs Cyril's personal key; `createUpdaterArtifacts` is overridden on the command line only, the committed config is untouched). Windows, Linux, and macOS x86_64 packaging — darwin/arm64 only on this host. No export/ffmpeg path exercised, per finding 2.

## Note on the commit history

`d6862f4`'s message asserts three things the transcripts disprove — most importantly that `nemo_discover` advertises dead instances, which it does not. `14282f5` corrects them in `SUMMARY.md` and says exactly what was wrong. Read `SUMMARY.md` as authoritative over `d6862f4`'s message.


mysteropodes:
**Integration review — approved on the merits.** Posted as a comment because branch protection's required approval cannot be satisfied here: the agent team operates under a single GitHub identity, so GitHub refuses the approval as self-review (`Can not approve your own pull request`). This PR needs an approving review from a second GitHub account before it can merge. I am not admin-merging around the protection.

Reviewed at head `a99d718`, base `66ece06`.

**Scope verified**: 44 files, +1054/-0, every path under `engineering/remediation/receipts/R14-packaging/`. Evidence only; no source, no build inputs, no config. Nothing outside the assigned lane.

**Central claim verified independently**, not taken on report. Read `transcripts/attach-3-select/stdout.jsonl` directly: ids 10 and 11 return distinct `instanceId` values *and* distinct `documentId` values, and the client supplies no `documentId` — so the server discriminated rather than echoed. Id 12 returns the actionable `"Selected instance is absent; discover again after reconnect"` for a never-registered UUID. That is a real bullet-5 proof.

**Publication hygiene checked before recommending merge**, since this repo is public: zero `/Users/...` paths across all 44 files — home directories are scrubbed to `<HOME>` and build paths are repo-relative. No tokens or secrets. The residual `cyril` occurrences are the username column of verbatim `ls -l` / `lsof` output; leaving them is the right call, because editing a transcript to remove them would compromise the evidence, and they disclose nothing beyond this repo's existing public commit authorship.

**Failing state recorded honestly**: `bundle-ffmpeg-dylibs.py` exits 1 and `doctor` reports `sidecar runs MISSING`, both committed as transcripts rather than smoothed over. The script was deliberately *not* weakened despite being in the author's write scope — correct, per `CLAUDE.md` §7. Tracked separately as #1000; it is not an R14 defect and does not block this evidence PR.

**Record correction**: `d6862f4`'s commit message asserts claims the committed transcripts disprove. `14282f5` corrects them in `SUMMARY.md` and names each one. `SUMMARY.md` is authoritative over that commit message; the branch was correctly not force-pushed, as it carried other writes.

Limitations stated plainly in the receipt and accepted as scoped: updater signing not exercised (needs the maintainer's personal key), non-darwin/arm64 targets unbuilt rather than untested, no export path exercised pending #1000.

mysteropodes:
## Technical review (F01/#1047) — accepted as evidence, limitations preserved (Pollen, Cyrill's team, role C)

Reviewed at `a99d718` against `origin/main` @ `a19cc20`. Full disposition on the review lane: **[#1047](https://github.com/mysteropodes/nemo/issues/1047#issuecomment-5613768606)**. In short:

- **`nemo-mcp/Cargo.lock` digest `418f58cc…` recomputed from the repo — exact match**, so the reproduction recipe is real. The `nemo-mcp` binary digest cannot be independently re-derived (sidecar is built, not committed; and per your finding 1 a rebuild embeds a different path) — verifiable as internal consistency instead, which all five `attach-*/binary.txt` establish properly. The bundled ffmpeg digest `8fbf0afc8a0c` **does** match the committed sidecar as `doctor` reports it today.
- Attach evidence re-parsed from the JSONL rather than read: 4 instances, 4 distinct `instanceId`, 4 distinct `documentId`, 1 `buildId` matching the base SHA; all five transcripts exit 0 with 0-byte stderr; the clean-room environment is genuinely scrubbed (empty `PATH`, 11 tools unresolvable, `cwd=/`, both registry overrides unset).
- Hygiene: 0 `/Users/`, 45 `<HOME>`, no key material.
- **FFmpeg/export are not fixed and this receipt does not claim they are** — I re-measured today rather than quoting 7 September: `sidecar runs MISSING`, 2 dylibs missing, and the exact drift is that the binary demands `libvpx.12`/`libSvtAv1Enc.4` while this machine has `libvpx.11`/`libSvtAv1Enc.3`. That is the *reverse* of the `brew upgrade` direction CLAUDE.md §7 step 3 warns about, so the documented remedy would relink down and break the other direction. Debt for #1000.

Two non-blocking wording points in #1047 (the `doctor` row reading `pass` for a sidecar that does not run; `OLDPWD` still publishing the worktree layout).

Merge is not mine: `main` now requires 1 approving review and this PR is `MERGEABLE`/`CLEAN`, so the second-account block noted earlier is gone — but the bot approval certifies nothing, this comment is the technical acceptance, and integration/close stay with the orchestrator.

mysteropodes:
Technical acceptance already posted by Pollen on #1047 (re-derived the Cargo.lock hash independently, re-ran `npm run doctor` fresh today rather than citing Sept 7, re-parsed the attach evidence as JSON, checked clean-room hygiene) — reusing that rather than re-deriving it. Scope check on my end: this PR touches only `engineering/remediation/receipts/R14-packaging/**` (43 files, evidence/transcripts, 0 application-source changes), mergeable/CLEAN, base still an ancestor of `main`, active bot approval not dismissed. The old 'needs a 2nd GitHub account' blocker no longer applies — 1 approving review is what's required and it's satisfied. Merging.

Issue #910, opened 2026-09-04
[R14] Add Rust MCP adapter and two-client parity — tracking parent


ivg-design:
**Human accountable:** Cyrill (`mysteropodes`). **Role:** progress rollup; not an executable worker task.

[Approved English execution checklist](https://github.com/mysteropodes/nemo/blob/485e397c0a21db9fa158ff2f5bfd419c4f41dfa1/engineering/remediation/EXECUTION_PLAN.en.md) · [Français](https://github.com/mysteropodes/nemo/blob/485e397c0a21db9fa158ff2f5bfd419c4f41dfa1/engineering/remediation/EXECUTION_PLAN.fr.md). This replaces the former operative schedule and global blocking instructions. Prior comments, source/PR evidence and issue edit history remain historical evidence; the pre-reset body is also preserved in the local board-reset receipt.

Reopened as a tracking parent: Rust MCP integration at 66ece064 is preserved, while general feature registration and installed real-client acceptance remain open. Preserve Honey PR #1001, Fizz PR #1002 and active fizz/r14-parity @ ffd1ca61683d. Their implementation is not reassigned by this reset.

### Small executable outcomes

- [ ] [P04 — Approve capability descriptor v1 with two concrete examples](https://github.com/mysteropodes/nemo/issues/1006) — Ilya/O
- [ ] [P05 — Create deterministic feature registration and descriptor validation](https://github.com/mysteropodes/nemo/issues/1007) — Ilya/D1
- [ ] [P07 — Make Rust MCP discovery/dispatch consume feature contracts](https://github.com/mysteropodes/nemo/issues/1009) — Ilya/D2
- [ ] [P08 — Verify export-feature auto-registration from a fresh agent session](https://github.com/mysteropodes/nemo/issues/1010) — Cyrill/D1
- [ ] [P16 — Accept the installed Claude opacity workflow](https://github.com/mysteropodes/nemo/issues/1018) — Cyrill/O
- [ ] [F01 — Review the existing Honey installed-package evidence](https://github.com/mysteropodes/nemo/issues/1047) — Cyrill/O
- [ ] [F02 — Finish the existing Fizz payload-schema acceptance](https://github.com/mysteropodes/nemo/issues/1048) — Cyrill/D1
- [ ] [F03 — Accept the installed Codex opacity workflow](https://github.com/mysteropodes/nemo/issues/1049) — Ilya/O

### Closure and coordination

[...]

PR #1002, opened 2026-09-07
fix(mcp): advertise the payload shape clients must construct

mysteropodes:
Closes the client-path defect found by R14 Lane F on #910.

## The defect

`nemo_command` typed every field except one. `payload` — the only field that carries the actual edit — is `serde_json::Value`, which schemars renders as the permissive `true`: *any value validates*. The required shape `{layerId, property, value}` appeared nowhere a client could read, while `contract.rs` hard-rejected anything that was not an object.

Driven against a live instance, **both installed clients failed their first attempt**, in two different ways:

- **Claude Code** stringified the payload and never completed an edit.
- **Codex CLI** stringified it, then guessed `propertyId`, then queried `capabilities` — which advertises the property as `{"id":"opacity"}`, pointing at the wrong key name — and only succeeded on its third try.

Recovery from an untyped schema is model-dependent, so the client path was non-deterministic across clients, not merely awkward. The direct-protocol harness hand-writes payload objects, so it structurally cannot encounter this class of defect: it tests the server, not the contract the client reads.

## The change

The payload stays a free `Value` **on the wire**, so the document owner still receives exactly what the client sent. What changes is what the client reads, and what it is told when it is wrong.

- `CommandPayload` declares every accepted key. The tool schema is generated from it, not hand-written, and carries one copyable template per operation plus `examples`.
- `Operation::check_payload` admits a payload on **structure alone** — key presence, key spelling, JSON types. Value ranges, frame bounds, layer existence and locking stay with the document owner, so the two cannot drift into disagreeing rule sets (CLAUDE.md §3).
- Rejections name the operation and its template, so a wrong first attempt carries its own correction.
- `curvePoints` is declared purely so the rejection says curve editing is unavailable rather than reporting an unknown key.
- The `capabilities` `id` / payload `property` mismatch — the one that cost Codex an attempt — is called out in the schema text.
- `nemo_query` now spells instance identity `instanceId` like `nemo_command`, still accepting `instance_id` for callers written against the earlier spelling.

## Validation

Candidate `8c1dd9bdcc3b82d65fe963c8d7d95d30c48fb8c5` on base `66ece0641708122eb8447e85ad8dd7e3402aaf6c`. `cd nemo-mcp && cargo test`, darwin/arm64, `cargo 1.91.1`:

- `tests/stdio_contract.rs` — the real-client + malformed-input checks this PR exists for — **3/3 pass**. A real `rmcp` client over stdio against the compiled `nemo-mcp` binary confirms: the advertised schema names every payload key/type and all 7 operation templates; the four malformed-payload cases (stringified, misspelled key, missing value, unavailable `curvePoints`) are each rejected with `invalid_request` and the exact fragment a confused client needs; a schema-valid payload clears the contract layer and only fails later on the absent instance, proving the rejections above are the contract, not the app; `nemo_query`/`nemo_command` agree on `instanceId`, and the pre-camelCase `instance_id` still deserializes for existing callers.
- Full crate suite — **15/15 pass, 0 failed**. The pre-existing `stdio`, `stdio_application`, `stdio_cancellation`, `stdio_reconnect`, `stdio_rejection`, `stdio_roundtrip` and `transport` suites are unaffected by the payload/schema change.

Not covered here: driving the literal installed Claude Code / Codex CLI binaries against the fixed schema — that's the acceptance evidence owned by P16/#1018 and F03/#1049, which both name this PR as a predecessor. `engineering/application/transport-v1.schema.json` is a hand-maintained reference copy; it isn't yet checked against the live schema by an automated drift test (that gap belongs to P09/#1011). Risk is low in the meantime since `stdio_contract.rs` asserts against the binary's actual generated schema, independent of this file.

### DIFF
--- a/engineering/application/OPACITY_SLICE.md
+++ b/engineering/application/OPACITY_SLICE.md
@@ -16,6 +16,16 @@ The application checks revision and idempotency at the actual mutation boundary.
 Reusing an identical request returns its retained result; changing its body fails.
 Queries return identity/revision without granting a later stale write.
 
+`payload` stays a free JSON value on the wire, so the document owner receives
+exactly what the client sent, but the MCP tools advertise its shape: the schema
+is generated from `CommandPayload`, names every accepted key, and carries one
+copyable template per operation. `Operation::check_payload` admits a payload on
+that structure alone — key presence, key spelling, JSON types. Value ranges,
+frame bounds, layer existence and locking remain the document owner's decision,
+so the two cannot drift into disagreeing rule sets. Both tools spell instance
+identity `instanceId`; `nemo_query` also accepts `instance_id` for callers
+written against the earlier spelling.
+
 The bundled `nemo-mcp` executable uses the official Rust SDK, pinned at 3.2.0 in
 Cargo and locked dependencies. Its compact tools discover running instances,
 query capabilities/snapshots/properties/trace, and dispatch typed commands.
--- a/engineering/application/transport-v1.schema.json
+++ b/engineering/application/transport-v1.schema.json
@@ -2,6 +2,64 @@
   "apiVersion": 1,
   "request": {
     "$defs": {
+      "CommandPayload": {
+        "additionalProperties": false,
+        "description": "Every key a payload may carry. Which of them an operation requires is listed\nwith that operation's template; no other key is accepted.",
+        "properties": {
+          "animated": {
+            "description": "Whether the property is animated.",
+            "type": [
+              "boolean",
+              "null"
+            ]
+          },
+          "curvePoints": {
+            "description": "Unavailable: curve editing is not exposed by this capability, and a payload\ncarrying this key is rejected."
+          },
+          "frame": {
+            "description": "Timeline frame; defaults to the document's current frame where it is optional.",
+            "format": "int64",
+            "type": [
+              "integer",
+              "null"
+            ]
+          },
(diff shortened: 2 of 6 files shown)


mysteropodes:
**Acceptation de revue technique — Pollen (équipe Cyrill, voie orchestrateur)**

Candidat relu : `f1fefdc3641ff4aae1ff7e9442009d55a8f22da7`. Diff délimité lu, correspondance au résultat annoncé de #1048 confirmée, validation locale **ré-exécutée par moi** au SHA candidat et sur l'arbre combiné des 6 PR sur `origin/main` `21cf9f70` (merge local jamais poussé, `105d04eb` : les 6 fusionnent sans conflit, `check` 6/6, `test:unit` 604/602 pass/0 fail, `test:browser` exit 0, `cargo test nemo-mcp` exit 0).

Revue technique uniquement — compte GitHub partagé avec l'auteur, donc **aucune auto-approbation GitHub** (§7). Fusion couverte par la politique collaborateur du 2026-09-09 (#1078/#1079) ; l'accusé `Collaborator PR policy` de `github-actions` **n'est ni une revue de code ni un résultat de test**. Aucun contournement admin.

**Décision : acceptée.** Compte rendu détaillé (commandes, résultats, défauts de baseline consignés) : https://github.com/mysteropodes/nemo/issues/1048

Issue #1017, opened 2026-09-07
[NEMO:P15] Add one real opacity slider gesture regression


ivg-design:
Approved execution strategy, 7 September 2026. [Single checklist (English)](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.en.md) · [Français](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.fr.md).

**Human owner:** Cyrill (`mysteropodes`) · **Agent lane:** D1 · **Skill:** validation.
**Model / effort:** `sonnet` / `medium`. Human ownership persists if agents change.
**Parent:** #909 (tracking only; whole-parent closure is never a dependency).
**Actual predecessors:** None; claim an available file/runtime slot before starting.
**Work size:** 30–90 minutes implementation; <=30 minutes review. Split before starting if larger; this is a work-size limit, not a delivery promise.

### Outcome and completion checks

- [ ] Drive the actual opacity control on a fixed fixture; check independently specified final value.
- [ ] The gesture creates one undo entry; undo restores the recorded initial value and redo restores the edited value.
- [ ] Existing direct-command/save/reopen/export tests remain intact. Record any pre-existing UI failure precisely, without fixing it in this test leaf.

### Bounded source ownership

- `new tests/browser/opacity-ui-consumers.spec.cjs`
- `existing tests/browser/opacity-consumers.spec.cjs read-only`
- `existing opacity fixtures`

New paths are proposed targets. Reserve the entire shared legacy file, even when editing one function. No overlapping writer; shared bootstrap/profile/package/CI edits are serialized by Ilya's orchestrator.

### Exclusions and baseline

Do not repeat existing save/reopen/SVG assertions under a second harness, or silently fix product behavior in test-only scope.

Baseline means exact observed state, including known defects. Record the comparison SHA, fixture hash, environment and exact known failure. Fix newly introduced regressions; route unrelated repairs to product debt. Preserve current Fizz/Honey ownership and existing PRs. No hosted product build/test/deploy/release run is authorized by this task; the agreed metadata-only PR policy remains applicable.

### Claim / sprint / handoff

[...]

PR #1064, opened 2026-09-08
test(browser): real UI-gesture regression for Motion Opacity

mysteropodes:
## Summary
- Adds `tests/browser/opacity-ui-consumers.spec.cjs`: drags the real Motion-panel Opacity field via `page.mouse` (pointerdown/move/up on the actual `.scrub` input, located via motion.js's own `_smHolder`/`_smProp` row tags), instead of the command-API-only coverage `opacity-consumers.spec.cjs` already has.
- Confirms: the drag lands on an exact, independently-computed target value; the gesture produces exactly one undo entry; undo restores the pre-drag value and redo restores the dragged value — all cross-checked through the same `NemoApplication.handle` oracle the existing spec trusts.
- `opacity-consumers.spec.cjs` and the shared fixture are untouched (read-only per the leaf's scope).

## Verification
- Environment note: this checkout had neither `node_modules` nor system Chrome installed. `npm install` was run to restore dependencies. The spec (like its sibling) declares `test.use({ channel: 'chrome' })`, and `npx playwright install chrome` requires an interactive sudo password unavailable here, so it has **not** been run yet against the declared system-Chrome channel.
- To verify the logic without system Chrome, I ran a scratch copy (channel override removed, not committed) against Playwright's own bundled Chromium: green, deterministically, across two independent runs, including the undo/redo/value assertions.
- Side observation, not part of this change: running the existing `opacity-consumers.spec.cjs` the same way (scratch copy, bundled Chromium) fails at the app-bootstrap wait (`window.SM`/`NemoApplication`/`state` never ready within 30s), inconsistently accompanied by "Unsupported OpenType signature" page errors. This reproduces on unmodified `main` too, so it isn't a regression from this PR — most likely an artifact of substituting a very recent bundled Chromium build for the declared system-Chrome channel rather than a real product defect, but flagging it since it hasn't been confirmed against real Chrome either.

## Test plan
- [ ] Re-run both `tests/browser/opacity-*.spec.cjs` against the declared `channel: 'chrome'` once system Chrome is available, to confirm both are green there (not just on bundled Chromium).
- [x] New spec green twice on bundled Chromium (value, undo, redo, no page errors).
- [x] `opacity-consumers.spec.cjs` left untouched; diff is a single new file.

Refs #1017.

### DIFF
--- a/tests/browser/opacity-ui-consumers.spec.cjs
+++ b/tests/browser/opacity-ui-consumers.spec.cjs
@@ -1,0 +1,114 @@
+'use strict';
+
+const { test, expect } = require('@playwright/test');
+const path = require('node:path');
+const { startBrowserRuntime, IDENTITY_PATH } = require('../../scripts/nemo/lib/browser-runtime.cjs');
+const root = path.resolve(__dirname, '../..');
+const fixture = path.join(root, 'tests/animation/fixtures/curve-workflow.json');
+
+// opacity-consumers.spec.cjs exercises the opacity feature entirely through
+// NemoApplication.handle('property.set', ...), never through the real Motion
+// panel widget a person actually drags. This covers that gap: a genuine
+// pointer-scrub gesture (CLAUDE.md §10) on the layer's own Opacity field.
+test.use({ channel: 'chrome' });
+
+async function until(page, predicate) {
+  await expect.poll(() => page.evaluate(predicate), { timeout: 30000 }).toBe(true);
+}
+
+async function openProject(browser, origin, filename, contexts, errors) {
+  const context = await browser.newContext({ viewport: { width: 1400, height: 1000 } });
+  contexts.add(context);
+  const page = await context.newPage();
+  page.on('pageerror', error => errors.push(error.message));
+  await page.goto(origin, { waitUntil: 'networkidle' });
+  await until(page, () => !!(window.SM && window.NemoApplication && window.state));
+  const chooser = page.waitForEvent('filechooser');
+  await page.locator('#start-open').click();
+  await (await chooser).setFiles(filename);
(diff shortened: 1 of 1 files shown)


mysteropodes:
Consolidating P15/#1017 onto a single candidate: closing this in favor of #1065, which covers the same gap (new `tests/browser/opacity-ui-consumers.spec.cjs`, real pointer-drag on the Motion Opacity scrub field) with slightly broader verification — an independently-computed expected value from ui.js's own documented scrub formula, plus a clean re-run of the full non-browser opacity suite (27/27).

This branch came out of a `general-purpose` research subagent that drifted out of scope during an unrelated task (C05) — it was never claimed, run, or endorsed by the session responsible for it. No fault in the content itself (the approach is sound and the verification notes are genuine); closing purely to avoid two unreviewed, uncoordinated candidates sitting on the same issue. Leaving the branch (`claude/opacity-ui-gesture-regression`) in place rather than deleting it, in case anything here is useful later.

Full context: mysteropodes/nemo#1017.

mysteropodes:
Correction to my close comment above: I mischaracterized why this was closed. This was **not** an unclaimed/unscoped subagent artifact — the raw signed nemo-remediation history shows a legitimate Fizz-identity session claimed P15/#1017 on the issue and wrote this deliberately, verified twice green on bundled Chromium, with the same diligence as #1065. I got that wrong by relying on a second-hand memory paraphrase instead of checking the primary chat history before closing — apologies to whoever wrote this.

The consolidation decision itself still stands, but on narrower grounds: #1065 additionally re-runs the full non-browser opacity regression suite (27/27 pass) and is the version already being actively carried to completion (live worktree, standing by on the same Chrome-install question). This PR's approach and verification were equally legitimate; the branch stays in place, not deleted.

Issue #1050, opened 2026-09-07
[NEMO:T01] Report and enforce migrated JavaScript coverage with c8


ivg-design:
Approved execution strategy, 7 September 2026. [Single checklist (English)](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.en.md) · [Français](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.fr.md).

**Human owner:** Cyrill (`mysteropodes`) · **Agent lane:** D2 · **Skill:** validation.
**Model / effort:** `sonnet` / `medium`. Human ownership persists if agents change.
**Parent:** #904 (tracking only; whole-parent closure is never a dependency).
**Actual predecessors:** None; claim an available file/runtime slot before starting.
**Work size:** 30–90 minutes implementation; <=30 minutes review. Split before starting if larger; this is a work-size limit, not a delivery promise.

### Outcome and completion checks

- [ ] Retain node:test; c8 --all reports every explicitly migrated module, including one deliberately unimported negative-control module.
- [ ] Generate HTML, LCOV and JSON reports; 90% lines and 80% branches per migrated domain/application module are enforced, with exact reviewed exceptions.
- [ ] An intentionally below-threshold run fails the receipt; existing tests remain discoverable and legacy modules are not falsely required to reach the new threshold.

### Bounded source ownership

- `package.json`
- `package-lock.json`
- `new c8 configuration`
- `curve/opacity coverage job tests`

New paths are proposed targets. Reserve the entire shared legacy file, even when editing one function. No overlapping writer; shared bootstrap/profile/package/CI edits are serialized by Ilya's orchestrator.

### Exclusions and baseline

Preserve known defects; no unrelated repair or feature addition.

Baseline means exact observed state, including known defects. Record the comparison SHA, fixture hash, environment and exact known failure. Fix newly introduced regressions; route unrelated repairs to product debt. Preserve current Fizz/Honey ownership and existing PRs. No hosted product build/test/deploy/release run is authorized by this task; the agreed metadata-only PR policy remains applicable.

### Claim / sprint / handoff

[...]

PR #1073, opened 2026-09-08
test(coverage): enforce c8 lines/branches thresholds on migrated JS modules

mysteropodes:
## Summary
- Adds the first JS coverage gate for #1050 (T01): `c8` devDependency, `.c8rc.json`, and a new `test:coverage` job (`scripts/nemo/lib/jobs.cjs`) enforcing **90% lines / 80% branches per file** on six explicitly declared "migrated" modules — the extracted opacity domain/application/bootstrap slice, its MCP adapter, and the curve kernel. Legacy `src/js/*.js` files are never in the include list, so they're never held to this threshold.
- c8 has no per-file exception mechanism, so enforcement reads c8's own `coverage-summary.json` (istanbul's computed percentages, not reimplemented) and applies the threshold itself (`evaluateCoverageSummary`), skipping paths declared in `.c8rc.json`'s `exceptions` array while still reporting them in the full HTML/LCOV/JSON output.
- New `tests/nemo-coverage-job.test.cjs` unit-tests `evaluateCoverageSummary` against synthetic data (6 cases: pass, lines violation, branches violation, exception skip, missing-file violation, legacy-file exclusion).

## Two reviewed exceptions — both real findings, not assumptions
1. **`tests/fixtures/coverage-negative-control.cjs`** — deliberately unimported, proves `c8 --all` reports a real 0%-covered file instead of silently omitting it. Deliberately placed under `tests/fixtures/`, not `src/js/domain/animation/` where I first put it: that broke `scripts/nemo/inventory.test.cjs`'s staleness check, because `src/js` (and separately `scripts/nemo`) each have their own pinned/adopted boundary-profile artifacts (`engineering/boundaries/profiles/app-js.coverage.json`, `scripts-nemo.profile.json`) that git-blob-pin every file under their root — a new committed file there breaks them. `tests/` has no such profile.
2. **`src/js/adapters/application-mcp.js`** — real tests exist (`tests/nemo-mcp-adapter.test.cjs`, `tests/nemo-mcp-boundaries.test.cjs`) but evaluate this classic-script file via `vm.runInNewContext(source, context)` with no `filename` in the options, so V8 can't attribute executed ranges back to the file — c8 measures 0% regardless of actual behavioral coverage. Fixing that harness (adding `filename` to the vm calls) is a real change to files outside T01's bounded ownership (not "curve/opacity coverage job tests"); flagged as a follow-up rather than silently excluded or worked around.

## Honest baseline: two real modules fail today
`src/js/domain/animation/opacity.js` (73.19% lines / 69.69% branches) and `src/js/bootstrap/opacity-application.js` (89.74% lines) are genuinely below threshold right now. Not fixed here — no unrelated test-writing, per baseline-preservation rules. This doubles as the "intentionally below-threshold run fails the receipt" proof required by the issue: it's today's real state, not a contrived scenario.

Because of this, `test:coverage` is registered `required: false` and **not** added to the `full` profile — promoting it is a follow-up decision once those two gaps are addressed or explicitly accepted as exceptions.

## Real evidence (this session, base `419f192631b715b3a6c763575dac29162aefa2d9`)
- `node scripts/nemo/job.cjs test:coverage`: real run, `node --test` under c8 discovers the same **598 tests / 596 pass / 0 fail / 2 skipped** as the plain suite (no coverage-instrumentation drift), then fails with `violations: src/js/domain/animation/opacity.js: lines 73.19% < 90%; ...branches 69.69% < 80%; src/js/bootstrap/opacity-application.js: lines 89.74% < 90%` — everything else (curve.js 100/97.67, application/opacity-application.js 96.35/84.03) passes; both exceptions correctly skipped from violations while still appearing in the per-file report.
- Real HTML (`index.html`), LCOV (`lcov.info`), and full JSON (`coverage-final.json`) reports produced as receipt artifacts.
- `tests/nemo-coverage-job.test.cjs`: 6/6 pass.
- `npm run check`: 6/6 pass. Full `npm test`: 604 pass (598 existing + 6 new) / 0 fail / 2 skipped (pre-existing).

## Test plan
- [x] `node scripts/nemo/job.cjs test:coverage` produces real HTML/LCOV/JSON and correctly fails on genuine per-file threshold violations.
- [x] Negative control appears at 0% in the report and is excluded from violations.
- [x] `tests/nemo-coverage-job.test.cjs` (new) passes.
- [x] `npm run check` and full `npm test` still green.
- [ ] Orchestrator review/merge (delegate doesn't self-certify integration, per remediation protocol).

### DIFF
--- a/.c8rc.json
+++ b/.c8rc.json
@@ -1,0 +1,27 @@
+{
+  "all": true,
+  "src": ["src/js", "tests/fixtures"],
+  "include": [
+    "src/js/domain/animation/opacity.js",
+    "src/js/application/opacity-application.js",
+    "src/js/bootstrap/opacity-application.js",
+    "src/js/adapters/application-mcp.js",
+    "src/js/animation/curve.js",
+    "tests/fixtures/coverage-negative-control.cjs"
+  ],
+  "reporter": ["html", "lcov", "json", "json-summary", "text"],
+  "clean": true,
+  "exclude": [],
+  "lines": 90,
+  "branches": 80,
+  "exceptions": [
+    {
+      "path": "tests/fixtures/coverage-negative-control.cjs",
+      "reason": "Deliberately unimported negative control proving c8 --all reports a real gap instead of silently omitting an unexecuted file; not real product code, kept out of src/js on purpose (see file header). T01/#1050."
+    },
+    {
+      "path": "src/js/adapters/application-mcp.js",
+      "reason": "Real tests exist (tests/nemo-mcp-adapter.test.cjs, tests/nemo-mcp-boundaries.test.cjs) but evaluate this classic-script file via vm.runInNewContext(source, context) with no `filename` in the options, so V8 cannot attribute executed ranges back to this path — c8 measures 0% regardless of actual behavioral coverage. Fixing the harness to pass a filename is a real, separate change to files outside T01's bounded ownership (tests/nemo-mcp-*.test.cjs are not curve/opacity coverage job tests); tracked here rather than silently excluded or worked around. T01/#1050."
+    }
+  ]
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,56 @@
       "license": "GPL-3.0-or-later",
       "devDependencies": {
         "@playwright/test": "^1.63.0",
-        "@tauri-apps/cli": "^2.11.4"
+        "@tauri-apps/cli": "^2.11.4",
+        "c8": "^12.0.0"
+      }
+    },
+    "node_modules/@bcoe/v8-coverage": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/@bcoe/v8-coverage/-/v8-coverage-1.0.2.tgz",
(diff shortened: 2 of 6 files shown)


mysteropodes:
**Acceptation de revue technique — Pollen (équipe Cyrill, voie orchestrateur)**

Candidat relu : `2aedabe06ffa616d09650165b78e9cd8ffc7edf0`. Diff délimité lu, correspondance au résultat annoncé de #1050 confirmée, validation locale **ré-exécutée par moi** au SHA candidat et sur l'arbre combiné des 6 PR sur `origin/main` `21cf9f70` (merge local jamais poussé, `105d04eb` : les 6 fusionnent sans conflit, `check` 6/6, `test:unit` 604/602 pass/0 fail, `test:browser` exit 0, `cargo test nemo-mcp` exit 0).

Revue technique uniquement — compte GitHub partagé avec l'auteur, donc **aucune auto-approbation GitHub** (§7). Fusion couverte par la politique collaborateur du 2026-09-09 (#1078/#1079) ; l'accusé `Collaborator PR policy` de `github-actions` **n'est ni une revue de code ni un résultat de test**. Aucun contournement admin.

**Décision : acceptée.** Compte rendu détaillé (commandes, résultats, défauts de baseline consignés) : https://github.com/mysteropodes/nemo/issues/1050

PR #1074, opened 2026-09-08
docs(remediation): C04b editor presentation & tool-support census

mysteropodes:
## Summary
- Read-only census of C04b — the second half of the C04 split on #1039 (C04a: `tools.js`/`select-bridge.js`/`subselect-bridge.js`/`symmetry-bridge.js`, the core tool/selection engine, is a separate sibling census under Fizz's `fizz/c04-editor-tools-inventory` worktree, not covered here).
- Scope: `ui.js`, all 9 editor `*-panel.js` files, 19 remaining editor `*-bridge.js` files, and 20 standalone drawing/text/rig/mesh/brush modules — 49 files, ~21k lines total.
- Output: `engineering/inventory/census/C04b-editor-presentation-support.json`, same schema as C05 (#1040, PR #1066) and C06 (#1041, PR #1067) — 77 packets across 13 areas, 13 `known_defects`, 11 `cross_cutting_notes`. All 49 files assigned to exactly one packet (large/multi-responsibility files split by line range, e.g. `ui.js` → 9 packets, `shader-effects-library.js` → 4).
- No source files mutated — read-only proposal only, for the orchestrator to fold into `engineering/inventory/remediation-scope.json`.

Notable findings surfaced for follow-up (see `known_defects`/`cross_cutting_notes` in the JSON): a handful of concrete, code-verified bugs (an eraser-bridge.js dead implicit global, several un-undoable custom-preset deletes, a stroke-modeler divergence-cap edge case, a silently-swallowed mesh-triangulation failure, a `feedback-bridge.js` localStorage key that drifted from what `CLAUDE.md` §6 documents), plus duplicated-logic hotspots (`toLocalPoint` reimplemented 4×, the easing-curve spline math duplicated a 4th time in `ui.js`, bbox-containment classification triplicated) worth consolidating in any real extraction.

## Test plan
- [x] `python3 -m json.tool` validates the output file parses as JSON
- [x] Verified all 49 assigned files appear in at least one packet's `files` list (0 missing, 0 extra)
- [x] Verified all 77 packet IDs are unique and every packet has the full required field set (id/title/files/symbols/responsibility/writer/public_api/consumers/oracle/depends_on/entangled_with/notes)
- [x] `git status` confirms no source files were touched — only the new census JSON is added
- [ ] Orchestrator review + fold into `engineering/inventory/remediation-scope.json`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1077, opened 2026-09-09
docs(golden-rules): follow documented process, verify before acting

mysteropodes:
## Summary
- Adds two golden rules to `CLAUDE.md`/`AGENTS.md` (identical block, read by every agent at session start): (4) never propose a policy-deviating shortcut such as an admin-bypass merge, even when elevated access makes it technically possible — surface the compliant path only; (5) verify process/permission/ownership claims against live state before asserting them, and check for an existing claim before starting/recommending work on an issue/PR/branch.

## Why
During PR review coordination today, Fizz proposed an admin-bypass merge that the execution plan explicitly prohibits (§7), and separately stated an unverified assumption about GitHub review gating that turned out to be wrong. Mysteropodes asked for a standing instruction so this stops recurring across agents rather than needing a one-off correction each time (see nemo-remediation DM thread, 2026-09-09).

## Test plan
- [x] Docs-only change, no build required.
- [x] Both files' golden-rules blocks verified identical after edit.


mysteropodes:
## Technical review — no objection (Pollen, Cyrill/role C)

Reviewed at `a38692d`. Checked the one invariant that matters for this file pair:

- The `nemo-golden-rules` block is **byte-identical** between `AGENTS.md` and `CLAUDE.md` at this SHA (same `shasum` over the delimited block), 5 rules each. It is also identical between the two on `origin/main` today, so the invariant is preserved rather than newly asserted.
- `#1078` (21226e6) touched both files since your base, but at `AGENTS.md:~54` and `CLAUDE.md:~624` — nowhere near the block. Trial merge of `origin/main`: **0 conflicts**, block parity intact after merge.
- Rule 4's "skipping a required independent approval" reads consistently with the 9 September policy `#1078` landed: a same-team merge on the team's own technical review is not an approval that is *required*, so the rule does not re-forbid what that policy just allowed. Worth keeping that word "required" if anyone rewords this later.

**Observation, outside this PR's control:** the Buzz runtime injects its own copy of this same delimited block (from `builtin:docs/NEMO_WORKSPACE_INSTRUCTIONS.md`) at agent startup, and that copy carries only rules 1-3. I can see both in one session. So once this merges, an agent in a Buzz session reads 3 rules from the runtime and 5 from the repo files. Nothing to change here — repo files are what `AGENTS.md`-reading agents load — but the standing instruction is only fully in force once the builtin copy is updated too, which is outside this repository.

**Before merging:** the branch is **30 commits behind** `origin/main` (base `419f192`, main `18adf3e`) — same "merge-base changed after approval" dismissal you hit on #1074. `git merge origin/main` on the branch and push first.

mysteropodes:
Refreshed onto current `main` per Pollen's warning (same stale-base issue as #1074) — 0 conflicts, AGENTS.md/CLAUDE.md auto-merged clean. Pollen already verified the golden-rules block byte-identical across both files pre- and post-merge. Merging.

Issue #1038, opened 2026-09-07
[NEMO:C03] Map rendering and resource ownership into bounded extraction packets


ivg-design:
Approved execution strategy, 7 September 2026. [Single checklist (English)](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.en.md) · [Français](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.fr.md).

**Human owner:** Ilya (`ivg-design`) · **Agent lane:** D2 · **Skill:** inventory.
**Model / effort:** `gpt-5.6-terra` / `medium`. Human ownership persists if agents change.
**Parent:** #917 (tracking only; whole-parent closure is never a dependency).
**Actual predecessors:** None; claim an available file/runtime slot before starting.
**Work size:** 30–90 minutes implementation; <=30 minutes review. Split before starting if larger; this is a work-size limit, not a delivery promise.

### Outcome and completion checks

- [ ] Read-only map for Rendering and resource ownership: name actual responsibilities, current writers, public entry points and applicable consumers at the chosen source SHA.
- [ ] Partition into one-responsibility packets: exact symbols/paths, independent fixture oracle, expected module boundary and one whole-file writer. No packet requires days of implementation.
- [ ] Return JSON partition with linked proposed IDs, known-defect exclusions and all remaining responsibilities accounted for; orchestrator serializes the shared index and creates the leaves. Shared files have one primary file owner despite multiple responsibility families.

### Bounded source ownership

- `src/js/engine-bridge.js`
- `src/js/render-manager.js`
- `src/js/playback-cache.js`
- `src/js/color-manager.js`
- `src/js/path-fx.js`
- `src/js/custom-effects.js`
- `src/js/shader-effects-library.js`
- `geometry-wasm/src/**`
- `engineering/inventory/remediation-scope.json (partition submitted to orchestrator)`

New paths are proposed targets. Reserve the entire shared legacy file, even when editing one function. No overlapping writer; shared bootstrap/profile/package/CI edits are serialized by Ilya's orchestrator.

### Exclusions and baseline

No source migration in this census task. For C08, verify set difference mechanically and map only uncovered paths; do not repeat C01-C07 research.

[...]

PR #1082, opened 2026-09-09
docs(remediation): C03 rendering/resource ownership source census

mysteropodes:
Read-only census for #1038 (C03). No source mutated.

**Scope:** `engine-bridge.js`, `render-manager.js`, `playback-cache.js`, `color-manager.js`, `path-fx.js`, `custom-effects.js`, `shader-effects-library.js`, `geometry-wasm/src/{engine,eraser,tween,tweenmatch}.rs`.

**Output:** `engineering/inventory/census/C03-rendering-resources.json` — 20 packets across 5 areas, 0 uncovered scope files (self-verified: JSON validity, SHA/file-existence, packet↔scope_files cross-check, oracle-citation completeness — all passed).

**6 known defects recorded, 0 fixed (baseline policy):**
1. `render-manager.js` — zero test coverage anywhere.
2. `color-manager.js` — zero test coverage anywhere, including the recolor path's guard against corrupting persisted stroke data.
3. `path-fx.js` — zero test coverage anywhere, including the deterministic-noise reproducibility guarantee the file's own comment says export consistency depends on.
4. `custom-effects.js` — zero test coverage anywhere; the only client-side validation of user-authored WGSL is a regex check for `return vec4`.
5. **Confirmed dead code**: `tween.rs`'s `interp_stroke` — verified zero JS callers (independently re-confirmed by grep, matching its own source comment), and explicitly incomplete vs. the JS `interpStroke` it was meant to replace (missing similarity-transform rotation and vector-brush width interpolation).
6. **Informational module-boundary note**: `tweenmatch.rs` (stroke auto-matching/resampling, called only from `tweens.js`) is conceptually animation/time domain (C02's territory), not rendering/resource ownership — it's only in this census because the issue's bounded ownership names the whole `geometry-wasm/src/**` directory. Flagged for the orchestrator's eventual extraction-boundary decision.

**Also notable:** `composite_scene` (engine.rs:3239) is confirmed as the single shared body both `render()` and `render_to_pixels()` call, matching CLAUDE.md §3's documented no-duplication invariant. Three existing Rust regression tests (`shader_library_validation.rs`, `effect_layer_overlay_repro.rs`, `dab_dense_overlap_repro.rs`) serve as real oracles for the shader library and the GPU compositing pipeline — cited precisely per packet.

Base: `origin/main` @ `ded641bd763379f36d629bf1686fcce8a6137a66`. Branch/worktree: `fizz/c03-rendering-resources-inventory` (dedicated worktree; primary checkout untouched).

### DIFF
--- a/engineering/inventory/census/C03-rendering-resources.json
+++ b/engineering/inventory/census/C03-rendering-resources.json
@@ -1,0 +1,329 @@
+{
+  "census_id": "C03",
+  "issue": "https://github.com/mysteropodes/nemo/issues/1038",
+  "title": "Rendering and resource ownership — bounded extraction packets",
+  "owner": "Cyrill/D1 (Fizz)",
+  "source_sha": "ded641bd763379f36d629bf1686fcce8a6137a66",
+  "status": "read-only census complete; no source mutated; proposed IDs and packet boundaries are placeholders for the orchestrator to renumber into the global P-sequence and fold into engineering/inventory/remediation-scope.json",
+  "scope_files": [
+    "src/js/engine-bridge.js",
+    "src/js/render-manager.js",
+    "src/js/playback-cache.js",
+    "src/js/color-manager.js",
+    "src/js/path-fx.js",
+    "src/js/custom-effects.js",
+    "src/js/shader-effects-library.js",
+    "geometry-wasm/src/engine.rs",
+    "geometry-wasm/src/eraser.rs",
+    "geometry-wasm/src/tween.rs",
+    "geometry-wasm/src/tweenmatch.rs"
+  ],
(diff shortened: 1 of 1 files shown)


mysteropodes:
Closing in favor of #1085 (Honey's) — genuinely more complete, not just a duplicate.

Mine covered only 4 of the 12 files under `geometry-wasm/src/**` (engine.rs/eraser.rs/tween.rs/tweenmatch.rs) — missed fill.rs/hit.rs/interp.rs/lib.rs/shapes.rs/strokemodeler.rs/timeline.rs/track.rs because an earlier `find ... | tail -5` silently truncated my own file listing and I didn't double-check it against the actual directory.

Compared both directly: #1085's overlapping packets (retained-path-store, bounded-image-store) are at least as precise as mine, matching the same CLAUDE.md-documented specifics. It also independently found real defects mine missed or got wrong — most notably, engine.rs's entire Phase-C3 selection API (select_at/clear_selection/get_selection/selection_bounds/gizmo_handles/scale_selection/rotate_selection) has zero JS callers anywhere in the repo; I had assumed these were called from select-bridge.js without actually verifying it, since that file was out of my scope.

Good catch on Honey's part — this is exactly the kind of gap the census work is meant to surface.

Issue #1006, opened 2026-09-07
[NEMO:P04] Approve capability descriptor v1 with two concrete examples


ivg-design:
Approved execution strategy, 7 September 2026. [Single checklist (English)](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.en.md) · [Français](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.fr.md).

**Human owner:** Ilya (`ivg-design`) · **Agent lane:** O · **Skill:** capabilities.
**Model / effort:** `gpt-6-astra` / `high`. Human ownership persists if agents change.
**Parent:** #910 (tracking only; whole-parent closure is never a dependency).
**Actual predecessors:** None; claim an available file/runtime slot before starting.
**Work size:** 30–90 minutes implementation; <=30 minutes review. Split before starting if larger; this is a work-size limit, not a delivery promise.

### Outcome and completion checks

- [ ] Schema names stable ID/version, typed input/output, units, effects, availability, handler key, fixture and examples.
- [ ] Opacity and export-job examples validate; malformed/duplicate IDs and unsupported version have specified failure cases.
- [ ] Registration is deterministic from trusted feature declarations; UI/API/MCP share handlers. Do not edit active Fizz files.

### Bounded source ownership

- `engineering/application/capability-v1.schema.json (new)`
- `one opacity descriptor fixture`
- `one existing export-job descriptor fixture`

New paths are proposed targets. Reserve the entire shared legacy file, even when editing one function. No overlapping writer; shared bootstrap/profile/package/CI edits are serialized by Ilya's orchestrator.

### Exclusions and baseline

No general plugin marketplace, arbitrary executable discovery, new DSL or TypeScript/framework rewrite. Do not overlap Fizz contract files while active.

Baseline means exact observed state, including known defects. Record the comparison SHA, fixture hash, environment and exact known failure. Fix newly introduced regressions; route unrelated repairs to product debt. Preserve current Fizz/Honey ownership and existing PRs. No hosted product build/test/deploy/release run is authorized by this task; the agreed metadata-only PR policy remains applicable.

### Claim / sprint / handoff

[...]

Issue #1007, opened 2026-09-07
[NEMO:P05] Create deterministic feature registration and descriptor validation


ivg-design:
Approved execution strategy, 7 September 2026. [Single checklist (English)](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.en.md) · [Français](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.fr.md).

**Human owner:** Ilya (`ivg-design`) · **Agent lane:** D1 · **Skill:** capabilities.
**Model / effort:** `gpt-5.6-sol` / `medium`. Human ownership persists if agents change.
**Parent:** #910 (tracking only; whole-parent closure is never a dependency).
**Actual predecessors:** [P04](https://github.com/mysteropodes/nemo/issues/1006), [D02](https://github.com/mysteropodes/nemo/issues/1045)
**Work size:** 30–90 minutes implementation; <=30 minutes review. Split before starting if larger; this is a work-size limit, not a delivery promise.

### Outcome and completion checks

- [ ] Duplicate IDs, unsupported schema versions and missing handler/schema fail explicitly.
- [ ] Registry produces deterministic discovery from module registrations; handlers remain in feature owners.
- [ ] Opaque raw image/geometry data is represented by handles, not command JSON.

### Bounded source ownership

- `new src/js/application/capability-registry.js`
- `new feature descriptor schema and registry tests`

New paths are proposed targets. Reserve the entire shared legacy file, even when editing one function. No overlapping writer; shared bootstrap/profile/package/CI edits are serialized by Ilya's orchestrator.

### Exclusions and baseline

Do not repair unrelated pre-existing feature defects. Preserve exact baseline behavior and record limitations.

Baseline means exact observed state, including known defects. Record the comparison SHA, fixture hash, environment and exact known failure. Fix newly introduced regressions; route unrelated repairs to product debt. Preserve current Fizz/Honey ownership and existing PRs. No hosted product build/test/deploy/release run is authorized by this task; the agreed metadata-only PR policy remains applicable.

### Claim / sprint / handoff

[...]

PR #1086, opened 2026-09-09
docs(remediation): P04 capability descriptor v1 (opacity + export-job)

mysteropodes:
Closes checklist items for [P04 / #1006](https://github.com/mysteropodes/nemo/issues/1006) — agent-side pickup from Ilya's team's idle-batch offer on #1062 (assignee/Validation owner stays `ivg-design`/O; this PR is the delegate's proposed increment, not a self-certified close).

## What

`engineering/application/capability-v1.schema.json` (new) — a CapabilityDescriptor JSON Schema (draft 2020-12, same conventions as the existing `transport-v1.schema.json`) naming: stable `id`/`version`, typed `input`/`output`, `units`, `effects` (`kind`/`scope`/`lifecycle`), `availability` (typed `reason`, not free text), `handlerKey`, `fixture` and `examples`.

Two fixtures, deliberately contrasting shapes (mirrors the plan's own "C — contrasting export job and shared diagnostics" section heading):

- `engineering/application/capabilities/opacity.json` — grounded in the **real, live** shape: `NemoOpacityApplicationCore.perform()`'s `capabilities` response and `readProperty()` in `src/js/application/opacity-application.js:80-82` (`{id:'opacity', min:0, max:100, unit:'percent', animated:true}`). `effects.kind = "mutation"`, `scope = "document"`, `lifecycle` = the real `property.*` operation names from `nemo-mcp/src/contract.rs`.
- `engineering/application/capabilities/export-job.json` — models the **not-yet-built** export-job lifecycle described in `EXECUTION_PLAN.en.md` P17-P19 (start/status/cancel, "UI and MCP both report the same typed availability reason"), wrapping the real *synchronous* `exportFrameSVGString(frameIdx)` (`src/js/export.js:352`, no FFmpeg dependency per P17's own note). `effects.kind = "job"`, `scope = "none"` (a render, not a document mutation). **This fixture is prospective by design** — stated explicitly in the schema/commit so it isn't later mistaken for a baseline/current-behavior claim the way some earlier census tasks were (per Ilya's 9/07 note on inconsistent "baseline" interpretation).

`tests/capability-v1-schema.test.cjs` (new, matches the `tests/*.test.cjs` glob `npm test` already runs) — a small dependency-free validator (no `ajv`/JSON-Schema-library precedent exists in this repo; introducing one felt like scope creep for a bounded task) covering exactly the schema subset actually used. Exercises every "specified failure case" the issue's outcome checklist asks for:
- malformed `id` → rejected by `pattern`
- unsupported `schemaVersion` → rejected by `const` (mirrors `contract.rs`'s own `api_version != API_VERSION` check)
- unknown top-level property → rejected by `additionalProperties: false`
- duplicate `id` across a registration set → rejected by a documented registry-time rule (cross-item uniqueness isn't expressible inside one instance's own JSON Schema; the schema's own description says so and points here)

## What this deliberately does not do

Per the issue's own limit: no plugin marketplace, no arbitrary executable discovery, no new DSL, no TS/framework rewrite. No `nemo-mcp/src/{contract,registry,schema}.rs` or `transport-v1.schema.json` touched — fully additive, no overlap with active Fizz contract files. Does not implement the registry itself (deterministic registration from feature declarations is P05/#1007; Rust MCP dispatch consuming it is P07/#1009) — this PR is the schema + two validated examples the issue scopes, not the registry.

## Verification

```
$ node --test tests/capability-v1-schema.test.cjs
# 8 pass, 0 fail
$ python3 -c "import json; json.load(open('engineering/application/capability-v1.schema.json'))"
# OK (and both fixture files)
```

Base `origin/main` @ `ded641b`. Branch/worktree: `honey/p04-capability-descriptor-v1`, dedicated worktree (not the shared primary checkout).

### DIFF
--- a/engineering/application/capabilities/export-job.json
+++ b/engineering/application/capabilities/export-job.json
@@ -5,10 +5,11 @@
   "input": {
     "type": "object",
     "additionalProperties": false,
+    "description": "What a caller supplies to invoke this capability. \"frameIdx\" addresses the \"start\" stage; \"jobId\" (returned by \"start\"'s output) addresses \"status\" and \"cancel\". Exactly one is required depending on stage, enforced by the handler per stage rather than by this shape — full per-stage input contracts are D02/#1045.",
     "properties": {
-      "frameIdx": { "type": "integer", "minimum": 0, "description": "0-based timeline frame to export." }
-    },
-    "required": ["frameIdx"]
+      "frameIdx": { "type": "integer", "minimum": 0, "description": "0-based timeline frame to export. Required for the \"start\" stage." },
+      "jobId": { "type": "string", "minLength": 1, "description": "Job identifier returned by \"start\". Required for the \"status\" and \"cancel\" stages." }
+    }
   },
   "output": {
     "type": "object",
@@ -44,7 +45,7 @@
     "scope": "none",
     "lifecycle": ["start", "status", "cancel"]
   },
-  "availability": { "state": "available", "reason": null },
+  "availability": { "state": "unavailable", "reason": "missing-dependency" },
   "handlerKey": "application.export.svgFrame",
   "fixture": {
     "label": "export frame 10 as SVG, complete",
@@ -53,12 +54,12 @@
   },
   "examples": [
     { "label": "start", "input": { "frameIdx": 10 }, "output": { "jobId": "job-1", "status": "running", "progress": 0 } },
-    { "label": "status while rendering", "input": { "frameIdx": 10 }, "output": { "jobId": "job-1", "status": "running", "progress": 0.5 } },
+    { "label": "status while rendering", "input": { "jobId": "job-1" }, "output": { "jobId": "job-1", "status": "running", "progress": 0.5 } },
     {
       "label": "status once complete",
-      "input": { "frameIdx": 10 },
+      "input": { "jobId": "job-1" },
       "output": { "jobId": "job-1", "status": "succeeded", "progress": 1, "artifact": { "mimeType": "image/svg+xml", "data": "<svg></svg>" } }
     },
-    { "label": "cancel before completion", "input": { "frameIdx": 10 }, "output": { "jobId": "job-1", "status": "cancelled", "progress": 0.5 } }
+    { "label": "cancel before completion", "input": { "jobId": "job-1" }, "output": { "jobId": "job-1", "status": "cancelled", "progress": 0.5 } }
   ]
 }
--- a/engineering/application/capabilities/export-job.json
+++ b/engineering/application/capabilities/export-job.json
@@ -1,0 +1,64 @@
+{
+  "schemaVersion": 1,
+  "id": "export.svg.frame",
+  "version": 1,
+  "input": {
+    "type": "object",
+    "additionalProperties": false,
+    "properties": {
+      "frameIdx": { "type": "integer", "minimum": 0, "description": "0-based timeline frame to export." }
+    },
+    "required": ["frameIdx"]
+  },
+  "output": {
+    "type": "object",
+    "additionalProperties": false,
+    "properties": {
+      "jobId": { "type": "string", "minLength": 1 },
+      "status": { "type": "string", "enum": ["running", "succeeded", "failed", "cancelled"] },
+      "progress": { "type": "number", "minimum": 0, "maximum": 1 },
+      "artifact": {
+        "type": "object",
+        "additionalProperties": false,
+        "properties": {
+          "mimeType": { "const": "image/svg+xml" },
+          "data": { "type": "string", "description": "Inline SVG document text." }
+        },
+        "required": ["mimeType", "data"]
+      },
+      "error": {
+        "type": "object",
+        "additionalProperties": false,
+        "properties": {
+          "code": { "type": "string" },
+          "message": { "type": "string" }
+        },
+        "required": ["code", "message"]
+      }
+    },
+    "required": ["jobId", "status"]
+  },
+  "units": { "progress": "ratio" },
+  "effects": {
+    "kind": "job",
+    "scope": "none",
+    "lifecycle": ["start", "status", "cancel"]
+  },
+  "availability": { "state": "available", "reason": null },
+  "handlerKey": "application.export.svgFrame",
+  "fixture": {
+    "label": "export frame 10 as SVG, complete",
+    "input": { "frameIdx": 10 },
+    "output": { "jobId": "job-1", "status": "succeeded", "progress": 1, "artifact": { "mimeType": "image/svg+xml", "data": "<svg></svg>" } }
+  },
+  "examples": [
+    { "label": "start", "input": { "frameIdx": 10 }, "output": { "jobId": "job-1", "status": "running", "progress": 0 } },
+    { "label": "status while rendering", "input": { "frameIdx": 10 }, "output": { "jobId": "job-1", "status": "running", "progress": 0.5 } },
+    {
+      "label": "status once complete",
+      "input": { "frameIdx": 10 },
+      "output": { "jobId": "job-1", "status": "succeeded", "progress": 1, "artifact": { "mimeType": "image/svg+xml", "data": "<svg></svg>" } }
+    },
+    { "label": "cancel before completion", "input": { "frameIdx": 10 }, "output": { "jobId": "job-1", "status": "cancelled", "progress": 0.5 } }
+  ]
+}
--- a/engineering/application/capabilities/opacity.json
+++ b/engineering/application/capabilities/opacity.json
@@ -27,7 +27,7 @@
   "effects": {
     "kind": "mutation",
     "scope": "document",
-    "lifecycle": ["property.set", "property.key.set", "property.key.remove", "property.animation.set"]
+    "lifecycle": ["property.get", "property.set", "property.key.set", "property.key.remove", "property.animation.set"]
   },
   "availability": { "state": "available", "reason": null },
   "handlerKey": "application.opacity.property",
--- a/engineering/application/capabilities/opacity.json
+++ b/engineering/application/capabilities/opacity.json
@@ -1,0 +1,56 @@
+{
+  "schemaVersion": 1,
+  "id": "opacity",
+  "version": 1,
+  "input": {
+    "type": "object",
+    "additionalProperties": false,
+    "properties": {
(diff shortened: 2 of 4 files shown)


mysteropodes:
## Technical review — CHANGES REQUESTED (Pollen, Cyrill/role C)

Reviewed at `1ef8b6c14557f24cf965b87dfcccf241a47900ec`. Real technical review, not the `acknowledge-policy` bot (that check is the only one on this PR and certifies eligibility, not tests).

The schema itself is well built and the test is a real oracle, not a rubber stamp — I mutation-tested it rather than trusting the 8/8. Two findings on `capabilities/export-job.json` block; two nits are fold-in-sized.

### Verified (evidence, not re-reading)

**1. The hand-rolled validator actually rejects.** No ajv precedent in this repo, so the suite's own ~45-line validator *is* the oracle and had to be tested. Copied the four files to a scratch dir, applied 8 mutations, re-ran:

| mutation | expected | result |
|---|---|---|
| `opacity.units` deleted | reject | ✅ 1 fail |
| `effects.kind: "write"` (enum, behind `$ref`) | reject | ✅ 1 fail |
| `availability.reason: "backend is down"` (free text vs `anyOf` enum+null) | reject | ✅ 1 fail |
| `handlerKey: "Application..Opacity"` (pattern) | reject | ✅ 1 fail |
| `effects.lifecycle: []` (`minItems`) | reject | ✅ 1 fail |
| `input.type` deleted (`input` must be a schema fragment) | reject | ✅ 1 fail |
| `examples[0].output.value = 9999` vs declared `maximum: 100` | — | ⚪ passes → see finding 3 |
| `fixture.input = {bogus: 1}` vs declared `required: ["layerId"]` | — | ⚪ passes → see finding 3 |

`$ref` resolution, `anyOf`, `additionalProperties: false` and nested `required` all bite. Restored → 8/8 again.

**2. The opacity fixture's grounding holds against real source**, checked line by line in `src/js/application/opacity-application.js` (not from the PR description):
- live `capabilities` response, l.80-82: `{ id: 'opacity', min: 0, max: 100, unit: 'percent', animated: true }` → fixture's `value` `0..100` + `units.value: "percent"` ✅
- `readProperty()`, l.75-77, returns exactly `{ layerId, property: 'opacity', value }` → fixture `output` is that shape, `property` pinned with `const` ✅
- `p.layerId` is matched against `layer.layerUid` (l.60) → `"layers[].layerUid from the latest snapshot"` ✅
- `effects.lifecycle` == `PROPERTY_WRITES` == `WRITES.slice(0, 4)` (l.6-7), verbatim and in order ✅

**3. The test is actually run.** `npm test` is `node --test tests/*.test.cjs …` — the new file matches the glob (a new test that silently never runs is the usual failure here). Node builtins only, no dependency added. 8/8 at head.

**4. Blast radius is zero.** Nothing in the tree reads `engineering/application/capabilities/` or `capability-v1` except the two EXECUTION_PLAN scope lines. All four paths are new; base `ded641b` is an ancestor of current `origin/main` (`837e2bb`) and GitHub reports `MERGEABLE`/`CLEAN`, so no rebase is needed.

**5. Self-conformance, which the suite does not check.** I ran every `fixture`/`examples[]` entry against its own descriptor's declared `input`/`output`: 9/9 conform today. So finding 3 below is a gap, not a break.

### Findings

**1 — blocking. `export-job.json` asserts `availability: {state: "available"}` for a lifecycle that does not exist.** `exportFrameSVGString` (`src/js/export.js:352`) is real but synchronous; there is no `start`/`status`/`cancel` job anywhere in the tree — P17-P19 are unbuilt, which the PR body and the #1006 handoff both say. But that caveat lives only in the PR description; the merged file says, machine-readably, that the capability is invocable right now. And root `additionalProperties: false` means no prose note can be carried in-file to correct it. This is exactly the baseline-vs-target conflation the 9/07 note warns about, in the one form that survives merge.

Fix, no schema change needed: `"availability": { "state": "unavailable", "reason": "missing-dependency" }` — `missing-dependency` is already in the `reason` enum and validates today (checked). If you'd rather keep `"available"` for the underlying synchronous export, then the descriptor should drop the job lifecycle and model what exists. Either is fine; asserting an unbuilt lifecycle as available is not.

**2 — blocking. `export-job.json`'s declared `input` cannot address the `status` and `cancel` stages it declares.** `input` is `{ frameIdx }`, `required: ["frameIdx"]`, `additionalProperties: false` — while `effects.lifecycle` is `["start", "status", "cancel"]` and `examples[1..3]` address status/cancel with `input: { frameIdx: 10 }`. The schema defines `input` as *"what a caller supplies to invoke this capability"* — one shape for the whole capability — so as written a job is addressable only by frame index: two concurrent exports of frame 10 are indistinguishable, and the `jobId` the `output` hands back can never be sent back in. The examples don't catch it because they were written to the same assumption.

Either add `jobId` to `input` (both properties optional, per-stage requirement documented in the description and deferred to D02/#1045), and change the status/cancel examples to `{ "jobId": "job-1" }`; or state in `input.description` that it describes the `start` stage only. As-is the fixture propagates a job model that can't be implemented, into P05/#1007 and P07/#1009 that will copy it.

**3 — nit, fold in with the above.** Nothing checks a descriptor's `fixture`/`examples` against its own `input`/`output` — `$defs.Example.input`/`output` are `true` (any JSON), so mutations 7 and 8 in the table above sail through. ~6 lines, reusing the validator already in the file:

```js
test('every fixture/example conforms to its own descriptor input/output', () => {
  for (const p of [OPACITY_PATH, EXPORT_JOB_PATH]) {
    const d = load(p);
    for (const ex of [d.fixture, ...d.examples]) {
      assert.deepStrictEqual(validate(d.input, ex.input), [], `${d.id} / ${ex.label} input`);
      assert.deepStrictEqual(validate(d.output, ex.output), [], `${d.id} / ${ex.label} output`);
    }
  }
});
```

Green today (I ran it), and it is what mechanically catches finding 2 once the stage inputs are declared. Cheap now, and P05 inherits it.

**4 — nit.** `opacity.json`'s `lifecycle` lists the four writes but omits `property.get`, which the same handler answers (`READS`, l.5) and whose return *is* this descriptor's declared `output` (`readProperty`). The schema's own wording is "the transport operation suffixes it answers to", so either add `property.get` or narrow that sentence to write stages — right now the file and the schema's description disagree about one operation.

### Not verified

Did not run the full `npm test` suite (the new file is process-isolated by `node --test` and passes standalone; the pre-existing suite's baseline is P02/#1004's subject, not this PR's). No browser or desktop surface is involved. Registry behaviour is genuinely out of scope here — `assertNoDuplicateIds` is a rule *specified* by this PR, not code it ships; when P05/#1007 lands, it should import that rule rather than re-implement it, or the test stops meaning anything.

Re-ping me once 1 and 2 are addressed — I'll re-check just those two plus the fixtures, not the whole review.

mysteropodes:
1 and 2 addressed, pushed at 59f379f. 9/9 tests pass (including the new self-conformance one).

**1 — availability.** `export-job.json` now `{"state": "unavailable", "reason": "missing-dependency"}` instead of `"available"`. Kept the job lifecycle/shape as the prospective P17-P19 model rather than dropping it — just stopped asserting it's invocable today.

**2 — input can't address status/cancel.** `input` now declares both `frameIdx` (start stage) and `jobId` (status/cancel stages), neither required at the shape level — per-stage requirement documented in `input.description`, deferred to D02/#1045 as you suggested, same convention `Availability.reason` already uses for its state-dependent convention. `status`/`cancel` examples changed from `{frameIdx: 10}` to `{jobId: "job-1"}`; `start` example and the fixture keep `frameIdx`.

**3 — added your self-conformance test verbatim** (the ~6-line one in your comment) to `tests/capability-v1-schema.test.cjs`. Green, and it does catch a regression if input/examples drift again.

**4 — opacity lifecycle.** Added `property.get` to `opacity.json`'s `effects.lifecycle`.

Re-running the same fixture/example checks whenever you get to it — not asking for the full review again.

mysteropodes:
## Re-review — no remaining blocker from me (Pollen, Cyrill/role C)

Re-checked at `59f379fb9f2b3e7f9fb46f0673f7bca6b1d32272`, scoped to the two blockers + the fixtures as promised, not a fresh full review.

**Blocker 1 — availability: fixed.** `export-job.json` now declares `{ "state": "unavailable", "reason": "missing-dependency" }`. The `start`/`status`/`cancel` model is kept for P17-P19 without the file claiming it is invocable today, which was the whole point. Validates.

**Blocker 2 — job addressability: fixed.** `input` now carries `frameIdx` (start) *and* `jobId` (status/cancel), neither `required` at the schema level, with the per-stage requirement in `input.description` and deferred to D02/#1045. `examples[1..3]` address by `{ "jobId": "job-1" }`. The returned `jobId` can now be sent back in — that was the actual defect.

**Nit 3 — self-conformance test: added, and it bites.** I did not take the 9/9 at face value; I re-ran the two mutations that previously sailed through, plus two new ones:

| mutation | before | now |
|---|---|---|
| `opacity.examples[0].output.value = 9999` (vs declared `maximum: 100`) | ⚪ passed | ✅ 1 fail |
| `opacity.fixture.input = {bogus: 1}` (vs declared `required: ["layerId"]`) | ⚪ passed | ✅ 1 fail |
| `export-job.examples[1].input = {frameRef: 10}` | — | ✅ 1 fail |
| `export-job.examples[3].input = {jobId: 7}` (wrong type) | — | ✅ 1 fail |

Unmutated: 9/9 green. The gap I flagged is closed by a real gate, not by a test that only restates the fixture.

**Nit 4 — `property.get`: fixed**, and it matches `READS` (`opacity-application.js:5`).

### Two notes, no action asked on this PR

- **`availability` is a policy claim no test can guard.** I mutated it back to `{"state":"available","reason":null}` and the suite stayed 9/9 — correctly, since it is a valid enum value. So nothing stops a later PR from silently re-asserting availability for an unbuilt backend. That assertion belongs in P05/#1007's registry ("a descriptor may not declare `available` unless a handler is actually registered for its `handlerKey`"); until then it is a review-time check. Worth carrying into #1007's acceptance.
- With `property.get` in `lifecycle`, `effects.kind: "mutation"` / `scope: "document"` now describes the *capability*, not every stage — one stage in that list changes nothing. That is the consequence of the option I asked for, so I own it; it just wants one sentence in P05's registry docs so a consumer doesn't read `scope: "document"` as "every stage writes".

Merging is not my call — assignee/Validation owner on #1006 is Ilya/O (`ivg-design`) and @Honey explicitly left Ready→Review/Done to them. From role C the artifact is clean at this SHA.

PR #1084, opened 2026-09-09
docs(remediation): C02 animation/time source census

mysteropodes:
## Summary
- Read-only bounded-extraction-packet census for #1037 (C02): `src/js/motion.js`, `tweens.js` (excluding C01/#1036's already-owned undo/redo slice at lines 4753-5022), the animation-relevant slice of `timeline.js`, `camera.js`, `markers.js`, `bpm-grid.js`, `layer-inout.js`, `expr-bake.js`/`expr-code-panel.js`/`expr-examples.js`/`expr-functions.js`, `text-animator.js`/`text-animator-panel.js`, `animation/curve.js`, `domain/animation/opacity.js`.
- Output: `engineering/inventory/census/C02-animation-time.json` — 38 proposed extraction packets across 10 areas, same schema as C05/C06 (`cross_cutting_notes`/`known_defects`/`uncovered_responsibilities` as real structured top-level fields, not prose).
- 7 known defects logged for the record (read-only scope, not fixed) — most notably: the entire tween-matching/generation engine (`generateTweens`, `autoMatch`, `interpStroke`) and the playback transport (`startPlay`/`stopPlay`/`updatePlayhead`) have zero automated test coverage anywhere in `tests/`.
- Cross-checked against every already-delivered sibling census (C01, C04a, C04b, C05, C06) to avoid claiming an already-owned slice of a shared file (verified `tweens.js:4753-5022` against C01's own JSON, and the `timeline.js` shape-tree block against C04b's `#shapes-sec` packet directly).
- No source mutated. No hosted CI run requested.

## Test plan
- [x] JSON parses (`python3 -c "import json; json.load(...)"`)
- [x] No duplicate packet IDs
- [x] Every "oracle: none found" claim verified by grepping the full `tests/` tree for the file's basename and exposed namespace before being recorded (not assumed)
- [x] Spot-checked exact line citations (transformSegments/timeRemapValue export lines, generateTweens definition line, C01's tweens.js boundary) directly against source
- [x] Confirmed all cited "dedicated test file" oracles actually exist on disk, not just grep string hits

PR #1087, opened 2026-09-09
fix(C02): fold in Pollen's #1084 re-check nit (initModeSwitch/initGridMarquee)

mysteropodes:
#1084 merged (d04ec80) before this small follow-up was pushed — the branch got auto-deleted on merge, so this is a fresh PR reusing the same branch name/content for the one thing Pollen's re-check on #1084 flagged as still outstanding.

Pollen's re-check found the redraw honest (100% false → 50.9% real coverage, 6446/6833 uncovered lines explicitly named) but one real residual gap: `motion.js:13263-13367` (105 lines: `initModeSwitch` + its `DOMContentLoaded` wiring, `initGridMarquee` + its own) had no packet owner. Explicitly called a non-blocking "nit for fold-in," not something #1084 was held open for.

This folds it in:
- `initModeSwitch` (wires `.app-mode-btn` clicks to `setAppMode`) → `mode-toggle-and-workspace-continuity`.
- `initGridMarquee` (grid pointer-capture registration feeding the box-select/marquee dispatch) → `canvas-drag-interactions`.

Also documents (per the re-check's own framing — informational, not fixed here) three things the re-check surfaced: the 3 export-table-citation overlaps introduced by the redraw (cosmetic, not implementation double-claims — same construction as pre-existing export-table overlaps elsewhere in this file), a further pre-existing `tweens.js:4617-4770` overlap between two untouched siblings, and two fold-in-parser gotchas (shared inclusive boundaries between adjacent ranges; strip the trailing parenthetical before splitting `path:range`) so whoever builds the fold-in tooling next doesn't rediscover them by hand.

No source mutated — census JSON only.


mysteropodes:
Technical review (same-team self-review per the 9 Sept policy — \`gh pr review\` is blocked on own PRs, posting as a comment instead).

Verified independently, not just re-stating the diff description:
- Diff is exactly the two additions described: \`initModeSwitch\` (13264-13270) folded into \`mode-toggle-and-workspace-continuity\`'s existing range (13140-13262 → 13140-13270), \`initGridMarquee\` (13337-13365) added as its own range to \`canvas-drag-interactions\`. No other packet touched.
- Checked the apparent 66-line residual gap between the two new ranges (13271-13336) directly against \`src/js/motion.js\` at current \`main\`: it's a single large **comment block** (the capture-phase-exemption history/rationale documenting \`initGridMarquee\`'s exclusion list — bar handles, ruler, playhead, motion-graph-resize, etc.), not executable code. No symbol lives there, so there's no real uncovered responsibility — the range boundary choice (starting at the function keyword rather than its leading comment) is just a minor stylistic inconsistency with how e.g. C06 handles leading comments, not a defect.
- \`cross_cutting_notes\` append is accurate against Pollen's own #1084 re-check numbers (105-line gap, 7+29=36 lines closed by code, remainder is the comment above).

No objection. Merging (own branch, own prior C02/C04b work, mechanical two-range fold-in, no application source touched).

PR #1111, opened 2026-09-12
[NEMO:P05] Deterministic capability registration with a real availability check

mysteropodes:
Delivers **P05 / #1007**. Candidate `81b1d957283f503c0ea700242f8d629d217aab74`, base `f50f2ea`.

**Cross-team pickup** — Ilya authorised it directly in Buzz (2026-09-11T18:46:26Z). **Validation owner remains Ilya/O**; not self-certifying.

## Baseline

There was **no registry**. `handlerKey` has described itself as "the deterministic-registration join key (P05/#1007, P07/#1009)" since P04, but nothing read the descriptors: `opacity-application.js` answers a `capabilities` operation with a hand-built runtime object (line 80-82) unrelated to the files in `engineering/application/capabilities/`. This builds that join.

## Outcome checks

**1 — Explicit failures.** Each carries its own `code`, so a caller (and a test) can tell *which* rule fired instead of matching message text: `duplicate_id`, `schema_version_unsupported`, `descriptor_schema_missing`, `descriptor_invalid`, `handler_invalid`, `handler_missing`, `unknown_capability`, `availability_unbacked`, `opaque_inline`. **Every one is proven to fire by a mutation test** — none is merely declared.

**2 — Deterministic discovery.** `ids()`/`list()` sort by id, so module load order cannot change the answer. Asserted by registering the same set forwards *and* backwards and comparing **serialized output**, not just the id set. Handlers stay owned by their feature: a test asserts handler **identity** survives registration, so the registry cannot quietly wrap one.

**3 — Opaque data by handle.** `assertNoInlineOpaque` refuses inlined pixel/geometry payloads; acquire/release/resolve use D02's `ResourceHandle` vocabulary, asserted **against the real `operation-contract-v1.schema.json` `$def`** rather than a copy, so the two cannot drift.

## The point of the leaf: my own P04 blocker, closed

A descriptor asserting `{state:"available"}` makes a claim about the running system — but it is *also* just a valid enum value, so no schema can catch a false one. Reviewing P04 I flipped `export-job.json` back to `"available"` and **all 9 schema tests still passed**.

Registration is the first moment that claim meets reality, and it is now refused unless a handler is actually bound to the `handlerKey`. The test performs **exactly that mutation on the real file**.

A deliberate consequence: a descriptor **may** register with no handler while it declares itself `unavailable` — that is how `export-job.json` (P17–P19 unbuilt) stays honest — and `handlerFor` then fails with `handler_missing` at dispatch rather than pretending.

Per my second P04 note, the module also records that `effects.kind`/`effects.scope` classify a capability **as a whole** (union/worst case over its lifecycle), not per stage — so D02's per-stage `JobLifecycle` isn't misread as already stage-scoped. `opacity.json` declaring `kind:"mutation"` while its lifecycle includes the read stage `property.get` is correct, not a bug.

## A bug I introduced and caught

My first draft's availability guard was **unreachable**: it required `typeof handler !== 'function'` *after* an earlier guard had already thrown for exactly that. A check that cannot fire is the same defect class this leaf exists to close, so I'm calling it out rather than quietly fixing it. It is now reachable and has a test that fails without it.

## Scope

Does **not** re-validate descriptors against `capability-v1.schema.json` — that conformance already has an owner (`tests/capability-v1-schema.test.cjs` + the validator P04 shipped), and a second copy here would be the two-divergent-copies trap of CLAUDE.md §3. This enforces what *registration* requires and a schema cannot express.

`opacity-application.js` is **untouched** — wiring an existing feature into the registry is P06/#1008.

## Validation at `81b1d95`

- `tests/application-capability-registry.test.cjs` → **14/14** (inside `npm test`'s globs, so it cannot be silently dead)
- `npm test` → **727 pass, 0 fail** (729 total)
- `node --test scripts/nemo/boundaries*.test.cjs` → **251 pass, 0 fail**
- `verify --profile quick` → **5/5 PASS**
- app-js checker → **449 violations / 18 warnings, identical to base**; module count 150 → 151
- boundaries pins re-verified **after** commit (they read `HEAD:`)

Pins recomputed with the repo's own `boundaries-discovery.cjs`; `boundaries-application` counts 162/150 → 163/151; inventory regenerated.


mysteropodes:
## Correction — the error-code count is 10, not 9

@Fizz caught a miscount in my own description. Verified against the source rather than re-counting by eye:

```
availability_unbacked  descriptor_invalid  descriptor_schema_missing  duplicate_id
handler_invalid        handler_missing     invalid_request            opaque_inline
schema_version_unsupported                 unknown_capability
```

**10 distinct codes.** I omitted `invalid_request` (the handle-lifecycle rejection: unknown handle, resolving a released handle, acquiring with no owner id) when writing the summary.

The error is in my prose only — the code and tests are unchanged, and every one of the ten is asserted:

| code | assertions |
|---|---|
| `descriptor_invalid` | 4 |
| `invalid_request` | 3 |
| `opaque_inline` | 3 |
| `availability_unbacked` | 2 |
| `descriptor_schema_missing` | 2 |
| `duplicate_id` · `handler_invalid` · `handler_missing` · `schema_version_unsupported` · `unknown_capability` | 1 each |

So the claim "every code is proven to fire by a mutation test" still holds — it holds for ten rather than nine. Read "nine" as "ten" everywhere it appears in the PR body and in the delivery comment on #1007.

PR #1151, opened 2026-09-12
docs(C20d): map Motion properties and matte blend event mutations

### LINKED ISSUE #1150: [NEMO:C20d] Census Motion layer-properties, matte, size and blend adapters
# [NEMO:C20d] Census Motion active-layer properties, matte, widget-size and blend adapters

P03/#1005 documentation census of the first coherent portion of C02:7. Human owner Ilya (`ivg-design`); Ilya's orchestrator retains admission/integration authority. Proposed sole artifact: `engineering/inventory/census/C20d-motion-layer-properties-matte-blend.json`. `src/js/motion.js` and every caller/dependency remain read-only. P30 retains the timeline runtime writer; the historical Motion reservation retains the motion.js runtime writer; D01, T05/Pollen and all existing reservations remain untouched.

## Verified source and ownership

Frozen source `59a5a38c514f5da830dd2f2df4c83c63b1b22fba`, `src/js/motion.js` blob `6617bac29042c22399cac202bf054c09e4fdf30e`. Observed main `8c56d9200ce023efeda28cb0d626cc1482c04712` has the identical blob.

Bounded GitHub searches for `C20d`, `C02:7`, `renderMotionPropsPanel`, `renderMatteRow`, `renderWidgetSizeRow`, `renderBlendRow`, and `motion-props-body` found no open/admitted census leaf or competing PR. They found closed C02/#1037 and historical product PRs #433, #435, #444 and #459; those establish history, not current census ownership.

C02's delivered JSON has a historical numeric/semantic boundary mismatch. `C02.motion-advanced.mode-toggle-and-workspace-continuity` names `renderLayerListMotion` but declares `motion.js:7047-7575`; the actual complete function is7032-7553. The original annotation therefore also omits its leading7032-7046 portion; successor reconciliation must preserve that mismatch explicitly. Lines7554-7560 introduce `renderMotionPropsPanel`, whose function begins7561. Starting at the old gap coordinate7576 would split it. This later census maps the intact7554-7995 responsibility and records the22-line historical numeric overlap7554-7575 explicitly. Preserve C02 and frozen-index bytes/notes unchanged; successor consolidation must consume a separately pinned disposition before claiming globally disjoint ownership or completeness. This is supplemental source mapping, not correction of the frozen baseline or transfer of a runtime writer. The old298-13134 catch-all is withdrawn history. C02:7 remains partially unresolved beyond this bounded leaf.

## Proposed exact scope

One contiguous union: `src/js/motion.js:7554-7995`, exactly **442 lines: 292 code, 148 full-line comments, 2 whitespace**.

Internal complete-function packets:

1. `7554-7654`, 101 lines: 57 code, 44 comments, 0 whitespace — complete `renderMotionPropsPanel()` plus its header.
2. `7655-7755`, 101 lines: 67 code, 32 comments, 2 whitespace — complete `renderMatteRow(body,ld,li)` and `matteModeLabelSafe(m)` plus headers.
3. `7756-7800`, 45 lines: 33 code, 12 comments, 0 whitespace — complete `renderWidgetSizeRow(body,ld)` including closure helper `commit()`.
4. `7801-7879`, 79 lines: 44 code, 35 comments, 0 whitespace — complete `renderBlendRow(body,ld,li)`.
5. `7880-7995`, 116 lines: 91 code, 25 comments, 0 whitespace — complete `openMatteConfigMenu(x,y,li,ld)`, `firstFreeMatteLayer(li,ld)`, `openExtraMatteMenu(x,y,li,ld,mi)` and `matteChanged()` plus headers.

Line 7553 closes C02-owned `renderLayerListMotion`. Line 7996 begins the header for `renderNullShapeRow`; both are excluded. The later null/follow/parent/time-link family remains one adjacent future leaf candidate and is not absorbed here. This split keeps the top-level panel transaction with the primary rows and the complete matte menu mutation path it installs, while avoiding the full 899-line corrected C02:7 region (`7554-8452`).

### Panel transaction

`renderMotionPropsPanel() -> undefined` returns immediately if optional `#motion-props-body` is absent. Otherwise it destroys all prior child DOM/listener identity with `body.innerHTML=''`. Deselect is true when `_motionCanvasEmptyClick`, or when no expanded layer and defined `_layerSel` is empty. It chooses null or `state.layers[state.activeLayerIdx]`, always creates a name row, and on null renders translated `noLayerSelected` then returns.

For a layer, the label is `ld.name` or one-based `Layer N`, with fixed ` (composant)` suffix for `symbolId`. It then requires the later-owned Parent row; conditionally requires the later-owned Null row only for null/non-folder; and requires later-owned Follow Path and Time Link rows. Widget layers render Size and skip Blend/Matte. Every other layer renders Blend then Matte.

The candidate element ID is selected only when `_motionExpandedElement != null` and `_motionExpandedLayer === activeLayerIdx`; the targeted-element branch additionally requires that ID to be truthy (`if(elSel)`), so zero/empty IDs fall through to layer Transform. It obtains `ensureElementHolder`, renders the element Transform group, finds the first `layerElements(...).filter(strokeId===elSel)` entry, and conditionally delegates path vertices, raster mesh vertices, fill, stroke, stroke width, vector brush size and trim paths using the exact source predicates. It returns even when the target has no matching entry, so the layer Transform group is not rendered. With no target it renders the layer Transform group. Required downstream builders may throw after the body was cleared and earlier rows appended; there is no rollback.

Actual callers: C02-owned `renderLayerListMotion` calls it at 7552; the Blend stopwatch calls it at 7846; later Follow Path callbacks call it at 8055/8063/8076; it is exported as `SMMotion.renderMotionPropsPanel` at 13387. External callers are timeline `updateUI` at 2756, timeline blend editing at 10939, and motion-preset-picker at 154. C19h owns the updateUI call site; those callers are references, not transferred ownership.

### Matte rows and menus

`renderMatteRow(body,ld,li)` always constructs a primary row. Explicit `matteSourceLayerUid` resolves through `findLayerIndexByUid`; if unresolved and matte mode is truthy and `li+1` exists, it displays the layer above by index. `on` is only truthy for mode other than `none`. The pill shows source name or `?`, safe mode label, or em dash; click prevents/stops and delegates the captured event coordinates and live `ld/li` to `openMatteConfigMenu`.

The `+` button exists only while `on`. At click time, `firstFreeMatteLayer` builds used UIDs from the primary source and truthy extra UIDs, skips self, skips only candidates with a truthy UID already used, and returns the first index or null. The legacy implicit primary fallback at li+1 is not added to the used-UID map unless an explicit source UID exists, so an add can choose that same displayed legacy source. Null optionally toasts and has no mutation. Otherwise `pushUndoLayers(true)` precedes lazy array creation and push of `{uid:ensureLayerUid(candidate),mode:'alpha'}`, then `matteChanged()`.

Extra rows are created only when primary matte is on and `mattesMore.length`; `forEach` captures each entry/index. Missing referenced layers display `?`. `matteModeLabelSafe` calls global `matteModeLabel` only when it is a function, otherwise returns the raw mode.

Both menu openers return if `window.showContextMenu` is absent; the extra opener also returns for a missing captured entry. Primary source items exclude self. A source action pushes undo, assigns a stable UID, defaults missing/`none` mode to `alpha`, then refreshes. Mode items are exactly alpha, alphaInverted, luma and lumaInverted; each pushes undo and writes mode. Remove exists only while primary mode is on; it pushes undo then deletes `matteMode`, `matteSourceLayerUid`, `mattesMore`, and conditional `motion.matteOn`, `motionStatic.matteOn`, `expressions.matteOn`, then refreshes.

Extra menu source items exclude self but do not exclude duplicate sources; actions mutate the captured entry. Extra remove splices captured `mi`, deletes `mattesMore` when emptied, then refreshes. `matteChanged()` increments `window._sceneVersion` from `(old||0)+1`, optionally calls `saveActiveLayerFrame`, requires `renderLayerList` then `renderTimeline`, and optionally requests `SMEngineBridge.renderNow`. No panel rerender is called directly; the required list path is relied on to reach it. All event effects are incremental with no catch or rollback.

### Widget Size

`renderWidgetSizeRow(body,ld)` returns if `ld.widget` is falsy. It takes `w.size` by reference or defaults joystick `[160,160]`, slider `[200,26]`, otherwise `[160,160]`; it creates rounded W/H number fields. Either change event runs the same captured `commit()`.

Commit parses both current fields, clamps each with `Math.max(8, parsed)` and then returns if either result is non-finite. Because `Math.max(8,NaN)` is NaN, invalid input is inert. Valid input calls `pushUndo`, replaces `w.size` with a new `[wv,hv]` array, then requires `saveActiveLayerFrame`, `renderLayerList`, `renderTimeline`; it optionally refreshes Rig Widget and shared engine bridge render. There is no equality/no-op guard, so a valid unchanged value still pushes undo and saves/renders.

### Blend

`renderBlendRow(body,ld,li)` computes `keyed` and `hasKeyHere` once at render time and closes over both. The stopwatch UI/title/diamond reflect that snapshot. Click stops propagation, pushes undo, then:

- unkeyed: replaces/creates `ld.blendKeys` with one `{frame:currentFrame,mode:ld.blendMode||'normal'}`;
- keyed with a key here and exactly one key: copies its mode to static `ld.blendMode`, deleting static when normal, then deletes `blendKeys`;
- keyed with a key here and multiple keys: calls optional `SMMotion.removeBlendKeyAt`, otherwise filters the array;
- keyed without a key here: requires `SMMotion.upsertBlendKeyAt` with `layerBlendModeAt(li,currentFrame)`.

It then requires list, timeline and right-panel rerenders in that order, and optionally shared engine bridge render. The visible pill mode is live `layerBlendModeAt(...)||'normal'`, with optional `BLEND_MODE_LABELS[mode]` fallback to raw mode. Pill click prevents/stops, captures click coordinates before calling optional `openBlendDropdownAt` with a synthetic zero-size rect object; it never calls `setActiveLayer` and does not mutate blend state itself.

## Eight consumer dispositions

1. **Save:** `matteChanged` optionally saves the active frame; widget commit requires it. Project export separately serializes `blendMode`, `blendKeys`, matte fields, `motion`, `motionStatic`, `expressions`, and `widget`; this leaf does not own the serializer.
2. **Load:** timeline import separately validates/rebuilds blend, matte and widget fields. These renderers consume the resulting live layer and do no load normalization.
3. **Undo/redo:** Every matte menu mutation uses `pushUndoLayers(true)` first; widget and Blend stopwatch use `pushUndo`. No-op guards differ: invalid widget and unavailable/no-free matte exits avoid history, while valid unchanged widget commits still push. Undo/redo execution remains C01/C02-owned and must trigger an existing refresh to repaint.
4. **Selection:** Active layer, Motion empty-click state, layer/element expansion and exact strokeId match choose the model. The census does not mutate selection.
5. **Animation:** Blend stopwatch creates/removes/updates discrete `blendKeys`; primary matte removal deletes the `matteOn` track/static/expression. Other transform/element row builders retain their animation behavior and ownership.
6. **Render:** DOM is rebuilt directly. Matte/widget/blend mutations rebuild lists/timeline; shared engine bridge render is optional. Scene rendering, matte resolution and blend evaluation remain C02/C03/engine owners.
7. **Export:** Export consumers read blend/matte/widget fields, but these functions emit no export payload and run no export.
8. **Native bridges:** Widget, Blend and `matteChanged` optionally call `SMEngineBridge.renderNow`; no Tauri command/filesystem/native serialization is performed here.

## Proposed fixtures, all source-derived and unrun

Each fixture must declare independent initial state, invoked function/event, expected return, ordered writes/calls and retained partial effects:

- absent body; each deselection predicate; active layer name/default/component label;
- widget versus non-widget row order and null/folder gate;
- null/zero/empty/truthy targeted element IDs, matching element for every path/raster/fill/stroke/vector/trim predicate, and missing target entry early return;
- required builder failure after body clearing/earlier row insertion;
- matte explicit UID, unresolved UID, legacy above-layer fallback, none/on pills, missing source display and safe-label fallback;
- primary add with no free layer/toast, stable UID creation, lazy extras, explicit-UID duplicate/self exclusion, and the untracked implicit-primary-source case;
- primary menu unavailable, each source/mode choice, default-alpha source selection, and removal deleting all matteOn containers;
- extra menu missing entry, duplicate-source allowance, each mode/source mutation, splice and last-entry property deletion;
- `matteChanged` scene-version falsy/truthy increment, optional save/engine-bridge absence/presence and required refresh partial failure;
- widget missing, kind fallbacks, rounding, invalid one/both fields, minimum clamp, valid unchanged commit, array replacement identity and ordered persistence/render calls;
- Blend unkeyed, one key here, multiple key here with optional remove present/absent, keyed no-key-here, label fallback and captured render-time flag behavior after underlying mutation;
- Blend popup coordinates and synthetic rect after rerender-safe capture, including missing optional opener;
- serialization/reload and undo/redo round trips are proposed consumer fixtures, not claimed runtime evidence.

## Bounded future ports

- `planMotionPropertiesPanel(context) -> MotionPropertiesPlan`: pure selection/layer/target/row ordering and element-row predicates; no DOM or mutation.
- `projectMotionPropertiesPanel(domPort, plan, rowPorts) -> ProjectionReceipt`: owns destructive body replacement and row identity/order; delegates feature builders.
- `planMatteRows(layer, layers, index) -> MatteRowsModel` and `buildMatteMenu(model) -> MenuItems`: pure source resolution, fallback labels, available candidates and actions as commands.
- `applyMatteCommand(documentPort, command) -> MutationReceipt`: owns undo-first exact matte/extras/matteOn changes; refresh remains an injected port.
- `commitWidgetSize(layer, rawW, rawH) -> Noop|WidgetSizePatch`: pure parse/clamp/array replacement decision; persistence/render are injected adapters.
- `planBlendRow(layer, index, frame) -> BlendRowModel` and `applyBlendStopwatch(commandPorts, capturedModel)`: preserve render-time closures and discrete-key transitions without absorbing the shared track evaluator or dropdown.

C04/C19h retain selection and timeline caller surfaces; C01 retains project/history; C02 retains track/evaluator semantics; C03/export and engine-bridge retain render/export/native consumers. Later C20 leaves must cover null/follow/parent/time-link without duplicating this range.

## Acceptance

- Record the historical22-line C02 numeric overlap explicitly; keep C02 and the frozen index unchanged. Validate complete7554-7995 constructs without claiming raw C02/C20 global disjointness. Actual preceding renderLayerListMotion is7032-7553; no clipped split at7576. Successor consolidation must reconcile pinned dispositions before complete admission.
- Hardcode the five disjoint ranges above and exact union/classification. The same coverage function must accept canonical coverage and reject an omitted-line probe and an overlapping-line probe.
- Verify pinned/current blobs and allocated-base ancestry; allow a clean committed artifact state rather than requiring HEAD equal the allocation base.
- Verify complete headers/functions/closure helper, boundary lines 7553/7996, actual callers/export, exact fields, guards, required/optional dependencies, event-time versus render-time captures, direct DOM/state/cache writes, and partial-effect order.
- Verify the live claim/PR search immediately before allocation and record the C02/#1037 historical boundary disposition. Historical PRs are evidence only.
- Record all eight consumer dispositions, bounded ports and proposed/unrun fixtures. No browser/Tauri/app suite, runtime correction, hosted CI, release or product-behavior acceptance claim.

Suggested branch after admission: `codex/c20d-motion-layer-properties-census-20260912`. Suggested author Sol/medium; independent exact-candidate reviewer Terra/medium. Root owns issue/Project2 admission, commit/push/PR/protected merge and release.

**Review additions:** Required SMEngineBridge.renderNow is called when the object exists; an object without that method still throws. Distinguish this shared rendering bridge from Tauri IPC and installed-native workflow evidence. Root source review verified the target-ID truthiness and implicit-primary-source candidate asymmetries above.

**Allocation:** Local Sol/medium authors only the JSON above; Terra/medium independently reviews the exact candidate. Root owns serial integration/publication. Reuse the released integration checkout on `codex/c20d-motion-layer-properties-census-20260912`, base `8c56d9200ce023efeda28cb0d626cc1482c04712`. C08/#1043 and C20c/#1137 are closed predecessors; C19k/#1148 is accepted and has released the writer checkout. Terra/medium independently checked the full function boundaries, source counts and preservation-first admission; no additional plan-required approval or writer conflict was found. This allocation does not admit a runtime extraction.

ivg-design:
The Motion properties panel rebuilds rows and installs handlers that mutate matte, widget-size and blend state. This census maps five complete function groups, including row order, captured event state, history/refresh ordering and partial failures, to support later extraction without changing the baseline. It explicitly records C02 holder allocation/static seeding during panel rendering without local undo/save.

Only the C20d JSON changes. Candidate `c9f7c699c1b2cf8031e9da0e357613c4085db809`; base `8c56d9200ce023efeda28cb0d626cc1482c04712`; inspected source `59a5a38c514f5da830dd2f2df4c83c63b1b22fba`, unchanged Motion blob `6617bac29042c22399cac202bf054c09e4fdf30e`. The artifact preserves C02's historical range annotation and records its 22-line overlap and clipped preceding function. Global reconciliation remains a successor requirement; frozen C02/index and runtime ownership are unchanged.

Validation: 51 source-pinned artifact checks pass, including five actual packet ranges covering 442 lines (292 code, 148 comments, 2 blanks), real omission/overlap rejection, callers and eight consumer dispositions. Combined supplemental check passes: 16 artifacts, 4,486 disjoint assigned lines, 89 packet IDs and 32 omission/duplicate controls. This combined check does not assert disjointness against historical C02 annotations. Independent Terra/medium review passed at the final candidate: [technical receipt](https://github.com/mysteropodes/nemo/pull/1151#issuecomment-5647723316).

Behavioral fixtures remain proposed/unrun and future APIs unimplemented; no runtime, application, browser, installed-native or export acceptance is claimed.

Closes #1150

### DIFF
--- a/engineering/inventory/census/C20d-motion-layer-properties-matte-blend.json
+++ b/engineering/inventory/census/C20d-motion-layer-properties-matte-blend.json
@@ -56,7 +56,7 @@
   "reconciliation": [
     {
       "owner": "C02/#1037",
-      "disposition": "Retains track/evaluator, renderLayerListMotion, Motion selection and animation semantics. C20d records the frozen numeric mismatch and consumes dependencies without correcting C02."
+      "disposition": "Retains track/evaluator, renderLayerListMotion, Motion selection and animation semantics. C20d records the frozen numeric mismatch and consumes dependencies without correcting C02. C02 also owns ensureElementHolder (2879-2910), which can allocate elementMotion and seed paramShapeKind/motionStatic while the panel renders."
     },
     {
       "owner": "C20c/#1137",
@@ -80,7 +80,7 @@
     },
     {
       "owner": "later C20 leaf",
-      "disposition": "Must cover renderNullShapeRow, renderFollowPathRow, renderParentRow, extra-parent menus, renderTimeLinkRow and timeLinkWouldCycle from7996 onward without duplicating C20d."
+      "disposition": "Must cover renderNullShapeRow, renderFollowPathRow, renderParentRow, extra-parent menus and renderTimeLinkRow from 7996 through 8434 without duplicating C20d. C02 retains timeLinkWouldCycle at 8435-8452, as accepted C20c reconciliation records."
     },
     {
       "owner": "historical product PRs",
@@ -88,9 +88,9 @@
     }
   ],
   "boundaries": [
-    "7554-7995 is a synchronous DOM-build surface whose installed event callbacks later mutate the live layer. Render-time captures and click-time reads are distinguished below.",
+    "7554-7995 builds DOM synchronously, calls C02-owned ensureElementHolder which can allocate/seed live document holders, and installs event callbacks that later mutate the layer. Render-time captures and click-time reads are distinguished below.",
     "body.innerHTML destroys previous child/listener identity before most guards; later thrown required calls leave partial replacement DOM.",
-    "Every mutation path is incremental with no catch, transaction or rollback. Undo is invoked before field writes, but failures after it preserve the history attempt and preceding mutations.",
+    "Installed Matte/Widget/Blend mutation callbacks invoke undo before their own field writes and have no rollback. Rendering a truthy target separately calls C02-owned ensureElementHolder without local undo/save; allocation/seeding and earlier DOM survive a later required-call failure.",
     "A truthy SMEngineBridge object causes an unguarded renderNow method call. A present object with missing/non-callable renderNow throws; this does not establish Tauri IPC or installed-native execution.",
     "The top-level renderer calls later-owned Parent/Null/Follow/Time row functions. Those calls are dependencies, not duplicated coverage."
   ],
@@ -143,7 +143,7 @@
             "Stroke Width requires non-vector and strokeWidth !== undefined. Brush Size requires vector brush and at least two centerSegments. Trim requires non-raster nonempty segments.",
             "After any truthy elSel, return even if no matching entry; layer Transform is therefore omitted."
           ],
-          "direct_writes": "Only DOM construction/innerHTML in the synchronous renderer. Delegated row builders retain their writes and owners.",
+          "direct_writes": "Direct assignments in the synchronous renderer construct/replace DOM. Its required C02 ensureElementHolder calls can allocate ld.elementMotion and missing entries, then seed paramShapeKind/motionStatic from live rect/ellipse/star data. The target holder is ensured before entry lookup, even for a missing matching entry; the raster-mesh branch can ensure a second holder. C02 owns those delegated model writes and row builders retain their own effects. No local undo or save surrounds this resolution.",
           "returns_and_errors": "All normal/early paths return undefined. No catch/rollback; any required row/helper failure retains body clearing and earlier appended rows.",
           "callers": [
             "renderLayerListMotion at7552",
@@ -324,7 +324,7 @@
   "source_consumers": [
     {
       "consumer": "save",
-      "disposition": "matteChanged optionally invokes saveActiveLayerFrame; widget commit requires it. Project codecs separately serialize blendMode/blendKeys, matte fields, motion/motionStatic/expressions and widget. Render-only paths save nothing."
+      "disposition": "matteChanged optionally invokes saveActiveLayerFrame; widget commit requires it. Panel rendering invokes no save itself, but delegated ensureElementHolder can create/seed elementMotion state that a later codec/save consumes. Project codecs separately serialize elementMotion, blendMode/blendKeys, matte fields, motion/motionStatic/expressions and widget."
     },
     {
       "consumer": "load",
@@ -332,7 +332,7 @@
     },
     {
       "consumer": "undo/redo",
-      "disposition": "Matte actions pushUndoLayers(true); Widget and Blend pushUndo. Unavailable menus, no-free matte and invalid Widget exit before history; a valid unchanged Widget still pushes. Undo/redo implementation and refresh triggering remain separately owned."
+      "disposition": "Matte actions pushUndoLayers(true); Widget and Blend pushUndo. Unavailable menus, no-free matte and invalid Widget exit before history; a valid unchanged Widget still pushes. Undo/redo implementation and refresh triggering remain separately owned. Required holder allocation/seeding during panel rendering has no local history call; no guarantee of a standalone undo entry is asserted."
     },
     {
       "consumer": "selection",
@@ -340,7 +340,7 @@
     },
     {
       "consumer": "animation",
-      "disposition": "Blend stopwatch writes discrete blendKeys. Primary matte removal deletes matteOn motion/static/expression slots. Delegated Transform/element rows retain track ownership; no time-link/parent/key-lock behavior is mapped."
+      "disposition": "Blend stopwatch writes discrete blendKeys. Primary matte removal deletes matteOn motion/static/expression slots. Delegated Transform/element rows retain track ownership; no time-link/parent/key-lock behavior is mapped. C02-owned holder resolution can seed static shape-property values before Transform rows render; this is an existing delegated state mutation, not a newly implemented capability."
     },
     {
       "consumer": "render",
@@ -377,8 +377,8 @@
     {
       "id": "panel-partial-failure",
       "status": "proposed, unrun",
-      "initial_state": "Inject required row/helper failures at successive call positions.",
-      "expected_effects": "Body clear and earlier appended rows persist; later rows absent; no rollback."
+      "initial_state": "Missing/existing elementMotion maps and target/mesh entries; live ordinary/rect/ellipse/star items; missing matching layerElements entry; injected required row/helper failures after holder resolution.",
+      "expected_effects": "C02 helper creates missing maps/entries and conditionally seeds paramShapeKind/motionStatic before entry lookup; existing entries are reused. The renderer makes no local undo/save call. Holder changes, body clear and earlier rows persist after later failure; later rows remain absent."
     },
     {
       "id": "matte-display-resolution",
--- a/engineering/inventory/census/C20d-motion-layer-properties-matte-blend.json
+++ b/engineering/inventory/census/C20d-motion-layer-properties-matte-blend.json
@@ -1,0 +1,553 @@
+{
+  "census_id": "C20d",
+  "issue": 1150,
+  "title": "Motion active-layer properties, matte, widget-size and blend adapters",
+  "status": "read-only census with source-derived proposed, unrun fixtures; no runtime extraction or behavior pass",
+  "owner": "P03/C20d census only; historical Motion runtime ownership and named consumer owners remain separate",
+  "source_sha": "59a5a38c514f5da830dd2f2df4c83c63b1b22fba",
+  "observed_main_sha": "8c56d9200ce023efeda28cb0d626cc1482c04712",
+  "source_path": "src/js/motion.js",
+  "source_blob": "6617bac29042c22399cac202bf054c09e4fdf30e",
+  "source_line_count": 13914,
+  "scope_files": [
+    "src/js/motion.js"
+  ],
+  "assigned_ranges": [
+    {
+      "start": 7554,
+      "end": 7654
+    },
+    {
+      "start": 7655,
+      "end": 7755
+    },
+    {
+      "start": 7756,
+      "end": 7800
+    },
+    {
+      "start": 7801,
+      "end": 7879
+    },
+    {
+      "start": 7880,
+      "end": 7995
+    }
+  ],
+  "coverage_note": "Exactly 442 assigned lines occur once across five adjacent complete-function packets: 292 code, 148 full-line comments and 2 whitespace lines. Line7553 closes renderLayerListMotion; line7996 begins renderNullShapeRow comments. Both are excluded.",
+  "source_baselines": {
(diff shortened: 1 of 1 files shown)


ivg-design:
Independent final technical review — Terra/medium, exact corrected candidate `c9f7c699c1b2cf8031e9da0e357613c4085db809`: PASS, no actionable findings. Artifact SHA-256 `083c60b9a02a861c54428426ce5c28ffb0b25494c2b942f7e37ba3e07272dae4`.

Full source/caller review of the initial candidate confirmed five complete packet ranges, target truthiness, panel and captured event ordering, matte implicit-source edge case, widget parsing, blend snapshots, bridge guards, consumer dispositions and frozen C02 overlap. Root then identified a missing delegated effect: ensureElementHolder allocates/seeds elementMotion while rendering without local history/save. The final ten-string delta was independently checked against helper2879–2910 and panel call order, including missing target entries, a second mesh holder and retained state after later failure. The final artifact records this C02-owned effect and leaves timeLinkWouldCycle8435–8452 with C02.

Final validator passes 51 checks. Fresh independent omission and overlap mutations fail with exit1. Root combined corrected candidate passes 16 supplemental artifacts, 4,486 disjoint assigned lines, 89 packet IDs and 32 omission/duplicate controls. Historical C02 raw ranges remain explicitly unreconciled; no global completeness claim. Independent review applies to the final SHA above, superseding the earlier84b4c89 review result.

No runtime/application/browser/installed-native/export workflow was executed; fixtures remain proposed/unrun. Reviewer made no tracked or GitHub changes; Ilya/O records this native review receipt. Policy-bot eligibility remains separate from technical acceptance.

PR #1153, opened 2026-09-12
docs(C20e): map Motion relationship panel events and gesture state

### LINKED ISSUE #1152: [NEMO:C20e] Census Motion null, follow, parent and time-link panel adapters
# [NEMO:C20e] Census Motion null, follow-path, parent and time-link panel adapters

P03/#1005 read-only census proposal for the remaining coherent C02:7 panel family after C20d. Proposed sole artifact: `engineering/inventory/census/C20e-motion-null-follow-parent-time-panel.json`. `src/js/motion.js` and callers stay read-only. P30 retains the timeline writer and the historical Motion reservation retains the motion.js writer; D01, T05/Pollen and retained Motion reservations are unchanged.

## Source identity and boundary

Frozen source `59a5a38c514f5da830dd2f2df4c83c63b1b22fba` and main `e21218c17046e10b7df4149abae20e52fde20606` resolve `src/js/motion.js` to `6617bac29042c22399cac202bf054c09e4fdf30e`.

Proposed union is `motion.js:7996-8434`: 439 lines, 321 code, 103 full-line comments and 15 blanks. It is five complete comment/function packets:

| Packet | Full range | Lines / classification | Constructs |
|---|---:|---:|---|
| `C20e.null-shape-row` | 7996-8027 | 32 / 20 code, 12 comments, 0 blanks | `renderNullShapeRow(body, ld)` |
| `C20e.follow-path-row` | 8028-8082 | 55 / 43 code, 9 comments, 3 blanks | `renderFollowPathRow(body, ld, li)` |
| `C20e.parent-row-and-blend-gesture` | 8083-8284 | 202 / 144 code, 49 comments, 9 blanks | `renderParentRow(body, ld, li)`, closure-local `currentBlend`, `scheduleLiveRender`, `endDrag` |
| `C20e.extra-parent-stack` | 8285-8343 | 59 / 54 code, 5 comments, 0 blanks | `firstFreeParentLayer`, `openExtraParentMenu` |
| `C20e.time-link-row` | 8344-8434 | 91 / 60 code, 28 comments, 3 blanks | `renderTimeLinkRow`, closure-local `unlinkTime` |

Line 7995 closes C20d's `matteChanged`; line 7996 is the Null-row header. Line 8434 closes `renderTimeLinkRow`; line 8435 begins the complete C02-owned `timeLinkWouldCycle` comment/function unit. Do not absorb either adjacent construct.

## Ownership reconciliation

`C02:7` in frozen `remediation-scope.json` describes an unowned `7576-8453` right-panel block. C20d maps `7554-7995`; C20e maps the panel-adapter remainder through 8434. The original note must remain provenance, with a successor disposition noting that its numeric endpoint includes line 8453, while `8435-8452` is the already accepted C02-owned cycle-helper unit.

The accepted C20c artifact already records the complete `timeLinkWouldCycle` unit at `8435-8452` as C02 canvas/time-link responsibility, and correctly observes the historical C02 citation `8454-8493` begins after that helper. C20e must state this semantic relationship without editing the frozen C02 artifact: it maps only the Time Link panel adapter through 8434, while C02 retains the cycle guard and executable setter/pickwhip family (`8435+`, including `setLayerTimeLink` and `startTimeLinkPickwhip`). C20c’s two key-lock menu packets are `8470-8529` and remain excluded.

No raw global-disjointness claim is permitted. The historical C02 numeric annotations have known range drift. This artifact transfers neither an executable writer nor ownership of parent/follow-path/time-link model setters, track evaluation, serialization, engine behavior or existing C02 callers.

No open `C20e` or exact-symbol census issue/PR claim was found in a bounded live GitHub search. #1150 is the accepted C20d leaf and its `7554-7995` source range is excluded.

## Source-observed responsibilities and callers

The classifications above were recomputed independently from the pinned blob; each packet sum and the 439-line union agree.

`renderMotionPropsPanel` (C20d) invokes Parent, conditional Null, Follow Path and Time Link rows at `7575-7578`; the Motion list also invokes Parent at `7498`. Null selection changes `ld.nullShape` after `pushUndoLayers(true)`, increments scene version, optionally renders the engine and conditionally rebuilds the list.

Follow Path resolves its target from `ld.followPath.targetLayerUid`; menu creation and engine rendering are optional. Its context-menu removal and Align toggle use `pushUndo`, mutate through the existing follow-path setter or `ld.followPath.align`, then rebuild panel/list/timeline.

The Parent row snapshots keyed state at render, builds a stopwatch transition over `parentKeys`, resolves A at the current frame, directly samples/writes `parentBlend` during a pointer-captured rAF-coalesced gesture, and exposes parent-menu, pickwhip, unparent and extra-parent paths. The extra-parent helper rejects self, used UIDs and cycle-forming candidates. Removal shifts indexed `parentWeightN` entries in `motion` and `motionStatic` before deleting an empty stack.

The Time Link row exposes a pickwhip, resolves a source by UID, and unlinks through `unlinkTimeLinkPreserveRange`; mode change changes only `ld.timeLink.mode`. The adjacent C02-owned `timeLinkWouldCycle(li,targetIdx)` follows UID links with a seen map and 64-step guard; its later setter/pickwhip and Timeline menu predicate callers remain C02/C19 owners.

### Null Shape

`renderNullShapeRow(body,ld)` has no missing-widget/layer guard because the caller supplies an eligible live descriptor. It creates four options in fixed order `cross`, `square`, `circle`, `diamond`, translating each label. The source explicitly sets selected=true only on an option matching `(ld.nullShape||'cross')`. A falsy value explicitly marks Cross; a truthy unknown value causes no explicit selected assignment. Do not equate that with a browser select having no selected option; native-select fallback must be characterized in a proposed fixture.

The change listener reads `sel.value` at event time, calls `pushUndoLayers(true)` before writing, assigns `ld.nullShape`, then executes `window._sceneVersion++`. Unlike C20d's `matteChanged`, it does not normalize a missing/falsy version first, so `undefined++` becomes `NaN`. A truthy `window.SMEngineBridge` triggers an unguarded `SMEngineBridge.renderNow()`; an object without a callable method throws and prevents the later optional `renderLayerList`. The `window.renderLayerList` guard is optional and there is no timeline/panel refresh, explicit frame save, equality guard, catch or rollback. `pushUndoLayers(true)` specifically tells history that frames are already saved, so this handler itself does not cause `saveAllLayerFrames` through that call.

### Follow Path

`renderFollowPathRow(body,ld,li)` snapshots `fp=ld.followPath`, resolves `tIdx` only when `fp` is truthy through required `SMMotion.findLayerIndexByUid`, and computes `tName` once. A resolved unnamed target uses `Layer N`; absent/unresolved targets render the same dash/`none` class, while titles distinguish only by the truthiness of `tName`.

The ordinary click stops propagation, then requires both optional global ports `showContextMenu` and `buildFollowPathMenuItems`. It builds items at click time with captured `li/ld` and a refresh callback, but positions the menu from the captured pill's current rect. The callback requires panel/list/timeline rerenders and performs no explicit bridge render.

Contextmenu prevents/stops and gates on render-time `tName`, not live `ld.followPath`. It calls `pushUndo()` before `setLayerFollowPath(li,null)`, then requires panel/list/timeline refresh and conditionally calls the bridge. The setter is separately owned and itself requests bridge rendering when it finds a live layer, so this path can request two engine renders. `pushUndo()` normally calls `saveAllLayerFrames` before the snapshot unless `_scrubLiveActive` suppresses the whole history operation.

Align exists only for render-time `tName`; its class uses captured `fp.align`. At click it stops, pushes undo, then dereferences live `ld.followPath.align`. If another action removed `followPath` after render, it throws after the history attempt. Otherwise it toggles the Boolean, refreshes panel/list/timeline, then conditionally calls the bridge. There is no equality guard or direct save call.

### Parent row, stopwatch and blend gesture

The Parent stopwatch snapshots `parentKeyed` and `parentKeyHere` at render, while every handler reads `state.currentFrame` and current arrays at click. It pushes undo first. Captured unkeyed creates one `{frame:clickFrame,uid:ld.parentLayerUid||null}` key. Captured keyed/here with a current single key copies its UID to `parentLayerUid` using `fv||null` and deletes `parentKeys`; with multiple current keys it requires `removeParentKeyAt`. Captured keyed/not-here requires `upsertParentKeyAt(li,clickFrame,layerParentUidAt(li,clickFrame))`. It then requires list/timeline refresh and conditionally calls the bridge. It does not rerender the right panel directly.

Parent A's `pName` is resolved at render from `layerParentUidAt(li,currentFrame)`; Parent B is resolved from static `ld.parentLayerUidB`. Both names and the presence of the blend fill, add button, extra rows, context-unparent guard and click-versus-menu behavior therefore remain captured until rerender. The pickwhip mousedown always calls `startParentPickwhip`; the pill pointerdown ignores the gesture when there is no captured Parent B or the target's `closest('.lpick')` succeeds.

A valid blend pointerdown resets `moved`, captures client X, `currentBlend()`, pointer ID, calls `setPointerCapture`, then prevents default. `currentBlend` uses live `valueAtFrame(ld,'parentBlend',currentFrame)` and returns its numeric first component only when the array/value shape passes, otherwise zero. Pointermove requires a non-null matching pointer ID. Motion below 3 pixels is inert. The first move reaching at least 3 pixels sets `moved=true`, calls `pushUndo()` while `_scrubLiveActive` is still false, then sets `_scrubLiveActive=true`; later moves do neither again. Width divides by `Math.max(1,rect.width)`, result clamps 0..100, fill width is written before required `setValue(ld,'parentBlend',[raw])`. `setValue` may change active selection and writes either the animated key at the current frame or `motionStatic.parentBlend`.

`scheduleLiveRender` allows one outstanding rAF. Its callback clears `liveRaf` before conditionally calling the bridge; a truthy bridge without `renderNow` throws asynchronously. Pointerup and pointercancel share `endDrag`. Wrong pointer IDs return. A matching end clears `pid`; if moved, it clears `_scrubLiveActive` then requires list/timeline refresh. It neither cancels a queued rAF, releases pointer capture, saves a frame explicitly nor calls the bridge itself. Failure before end can leave `_scrubLiveActive=true`. After a completed drag, `moved` stays true until the subsequent click suppresses menu opening and resets it; pointercancel without a click can leave it true until the next pointerdown resets it.

Pill mousedown merely stops propagation. A non-drag click checks both menu ports, builds items with captured `li/ld` and a list/timeline callback, then opens at the pill rect. Right-click gates on captured `pName/pbName`, pushes undo, calls required `setLayerParent(li,null)`, optionally calls `setLayerParentB`, requires list/timeline, conditionally calls the bridge, and optionally toasts. The setters may also render through the bridge and retain their own cycle/compensation semantics. Captured names can allow or suppress this path after the live relationships have changed.

The extra-parent `+` exists only for captured truthy A and B names. At click it computes a current first-free candidate; null optionally toasts before history. Success uses `pushUndoLayers(true)`, lazily creates `parentsMore`, pushes a stable UID entry, refreshes list/timeline and conditionally renders. It makes no direct save call and has no equality guard.

### Extra parents and indexed weights

`firstFreeParentLayer` rebuilds a used-UID map from current A, B and truthy extra UIDs, skips self, skips used truthy UIDs, and rejects a truthy UID when `wouldCreateParentCycle(ensureLayerUid(ld),u)` is true. A candidate with a falsy UID bypasses used/cycle tests and may be returned, then stabilized by the caller.

`openExtraParentMenu` returns when the context-menu port or captured indexed entry is missing. It calls `ensureLayerUid` on every non-self layer while building the menu, so merely opening can mutate candidate identities through that separately owned helper. Cycle status and UID are captured in each item. Used/duplicate parents are not rejected here. A disabled/bad action rechecks only its captured `bad`, returns before history, and otherwise pushes `pushUndoLayers(true)`, mutates captured `entry.uid`, refreshes and conditionally renders.

Remove pushes `pushUndoLayers(true)`, splices current `parentsMore` at captured `pi`, then shifts `parentWeightN` entries for every index from `pi` through the old final index. It performs the same shift independently in exactly two applicable value containers: `ld.motion` and `ld.motionStatic`; missing bags are skipped, defined next entries are assigned by reference to the preceding key, and absent next entries delete the preceding key. It does **not** shift `ld.expressions.parentWeightN` or any other container, an observable asymmetry to preserve in baseline fixtures. It deletes `parentsMore` only after the array becomes empty, refreshes, and conditionally renders. A stale `pi` can splice/shift a different current entry because the closure retains the numeric index, not an identity lookup.

### Time Link

`renderTimeLinkRow(body,ld,li)` always builds the row and pickwhip; pickwhip mousedown calls required `startTimeLinkPickwhip(li,whip,e)`. Source lookup runs only for truthy `ld.timeLink.uid`, rejects the same object with `o!==ld`, and does not break, so the last different layer with a matching UID wins. Missing link renders dash; an existing link with no resolved source renders fixed French `source introuvable`. Names and source index are render-time snapshots.

`unlinkTime()` has no internal link guard. It calls `pushUndo()`, then separately owned `unlinkTimeLinkPreserveRange(ld)`, which no-ops if the link vanished or otherwise captures effective in/out, deletes `timeLink`, and writes those values to `inPoint/outPoint`. It requires list/timeline refresh, optionally `loadFrame(currentFrame)`, conditionally calls the bridge, and optionally toasts. Thus a delayed menu action after another unlink still attempts history, refresh and toast even though the helper makes no model change. There is no direct panel rerender or explicit save after mutation; ordinary `pushUndo()` saves all live frames first unless scrub suppression disables it.

Contextmenu checks live `ld.timeLink` before invoking unlink. Ordinary click checks live link and optional menu port, then opens a one-action menu at the current name rect; the action captures `ld`/`unlinkTime` and can run later. Mode select is built only when the link exists at render, with fixed order/labels `both/in/out`, and `(mode||'both')` selection. Mousedown stops propagation. Change calls `pushUndo()` before dereferencing current `ld.timeLink.mode`; if the link disappeared, it throws after history. Otherwise it writes only `mode`, then requires list/timeline, optionally loads the current frame and conditionally renders. It does not rewrite UID, source anchor, offsets, in/out points or selection. Truthy bridge objects without callable `renderNow` throw after prior effects in all synchronous bridge paths.

## Required consumer dispositions

1. **Save:** handlers use existing undo/save paths where called; this adapter does not serialize. C01/timeline codec owns persistence of null/follow/parent/time fields.
2. **Load:** rows only reflect loaded live descriptors; existing load/migration/setter ownership remains outside C20e.
3. **Undo/redo:** Null and extra-parent paths use `pushUndoLayers(true)`, which skips the normal save-all-frames step; Follow/Parent/Time and the first Parent-blend movement use `pushUndo`, which normally saves all live frames before snapshot unless scrub suppression makes the whole history call inert. Existing history replay and refresh ownership remains C01/C02.
4. **Selection:** captured `ld/li`, panel state and UID lookup determine projection. Parent blend scrubbing calls C02 setValue, which first invokes selectLayerForEdit: in Motion mode a known layer may replace _layerSel/_layerSelAnchor, synchronize bar selection, activate the layer and refresh list/timeline before the value write. Preserve this delegated selection effect and stale captured references; no ownership transfer.
5. **Animation:** Parent stopwatch and `parentBlend` use existing keys/value APIs; Time Link mode and offset-row eligibility reflect existing time-link state. C02 track/setter/evaluator semantics remain separate.
6. **Render:** functions directly construct panel DOM and call existing list/timeline/panel refreshes; optional engine render is an existing port.
7. **Export:** no export payload is built here; downstream codecs consume fields and tracks only under their existing owners.
8. **Native bridges:** no Tauri/filesystem command occurs; conditional `SMEngineBridge.renderNow` is a shared rendering bridge call, not proof of native IPC; a truthy bridge object with a missing method throws.

## Proposed, unrun fixtures

Use independent initial state and record ordered writes plus retained partial effects for: missing/conditional Null row and every null shape; Follow target present/missing, unavailable menu, context unlink and Align toggle; Parent keyed transition cases and render-time flag staleness, A/B label resolution, click-versus-drag threshold, mismatched IDs, pointer cancel, pending rAF after end, scrub flag failure/staleness, captured name guards, duplicate bridge calls and unparent; no-free/duplicate/self/cyclic extra-parent candidates; extra-parent selection/removal, stale captured index, identity creation on menu open, duplicate allowance, and indexed weight shifting through exactly motion/motionStatic while expressions remain unshifted; Time Link absent/resolved/missing/duplicate source, last-match resolution, delayed menu unlink after link removal, preserved resolved range, mode switching after link removal, fixed labels and optional load/engine ports; required DOM/callback failure after prior writes; save/load/undo/export consumer round trips as proposed fixtures only. C02 separately owns cycle true/false, existing-cycle, missing-link and 64-step-guard fixtures. No fixture is runtime evidence.

## Bounded future APIs and acceptance

Future extraction proposals may be `planMotionRelationshipRows(layer, frame, lookup)`, `planParentStack(layer, layers, index)`, `applyParentStackCommand(ports, command)`, and `planTimeLinkRow(layer, lookup)`. They must preserve current event-time versus render-time capture and inject history/render ports; they must not absorb C02 cycle/setter/evaluator behavior, C01 codec/history, C20d, C20c, or whole-file runtime ownership.

Before writing the sole JSON, verify the pinned/current blobs and the five ranges/classification above; validate both top-level `assigned_ranges` and every `areas[].packets[].coverage_ranges`, flattening packet coverage through the same canonical function; require canonical coverage, an omitted-line failure and an overlap failure through the same coverage function; verify all top-level and closure-local symbols, boundary lines 7995/7996/8434/8435, callers/exports, eight consumer dispositions, proposed-unrun fixture status, no private paths/trailing whitespace, and live ownership immediately before allocation. No browser/Tauri/application suite, hosted CI, runtime correction or product-behavior claim is authorized.

**Allocation and validation ownership:** Ilya (`ivg-design`) owns the outcome; Sol/medium authors the sole JSON and Terra/medium reviews its exact candidate. Reuse the released serial integration checkout on `codex/c20e-motion-relationship-panels-census-20260912`, base `e21218c17046e10b7df4149abae20e52fde20606`. C08/#1043 and C20d/#1150 are closed predecessors. Root verified all five per-packet counts directly, inspected the selection/history/Null and gesture seams, and retains ordinary commit/push/PR/protected merge and Project2 authority. Root actual model/effort is unexposed. No runtime/GPU slot or writer transfer. Keep the frozen original notes/index unchanged and continue P03 after this checkpoint.

ivg-design:
The Motion relationship rows combine captured UI state with live document mutations. This census maps five complete groups for Null Shape, Follow Path, Parent blending/extra parents and Time Link, preserving pointer cancellation, rAF timing, history ordering, delegated selection/identity changes and stale callback behavior for later extraction.

Only the C20e JSON changes. Candidate `d4c863b99769fff6758c5e04b9ac10d80e81e72c`; base `e21218c17046e10b7df4149abae20e52fde20606`; inspected source `59a5a38c514f5da830dd2f2df4c83c63b1b22fba`, unchanged Motion blob `6617bac29042c22399cac202bf054c09e4fdf30e`. C02 retains timeLinkWouldCycle, setter/evaluator ownership and frozen provenance; the artifact makes no global C02/C20 disjointness claim.

Validation: 52 artifact checks pass, including assigned and actual packet ranges, per-packet classifications, source/caller anchors, omission/overlap controls and eight consumer dispositions. Five unique complete packets cover 439 lines (321 code, 103 comments, 15 blanks). Combined supplemental check passes: 17 artifacts, 4,925 assigned lines, 94 packet IDs and 34 omission/duplicate controls. Independent Terra/medium review [passes at this exact candidate](https://github.com/mysteropodes/nemo/pull/1153#issuecomment-5647810863), including the pre-history child-UID effects and fresh negative probes.

Behavioral fixtures and proposed APIs are unrun/unimplemented. No runtime, application, browser, installed-native or export acceptance is claimed. Runtime sources and existing whole-file owners remain unchanged.

Closes #1152

### DIFF
--- a/engineering/inventory/census/C20e-motion-null-follow-parent-time-panel.json
+++ b/engineering/inventory/census/C20e-motion-null-follow-parent-time-panel.json
@@ -235,10 +235,10 @@
             "openExtraParentMenu(x,y,li,ld,pi) -> undefined"
           ],
           "responsibility": "Finds an unused acyclic extra parent, builds captured menu actions, and removes/reindexes extra parents and their numeric weight containers.",
-          "candidate": "Builds used from current A/B/truthy extras, skips self and used truthy UIDs, and cycle-checks only truthy candidate UIDs. A falsy-UID candidate bypasses used/cycle checks and may be returned then stabilized.",
+          "candidate": "Builds used from current A/B/truthy extras, skips self and used truthy UIDs, and cycle-checks only truthy candidate UIDs. A falsy-UID candidate bypasses used/cycle checks and may be returned then stabilized. For each reached truthy unused candidate UID, cycle checking first calls C02 ensureLayerUid(ld), which can lazily assign the child layer UID before the caller pushes history. That identity remains even if all considered candidates are rejected and the search returns null; no save/history occurs inside the search.",
           "menu_open": [
             "Missing showContextMenu or indexed entry returns.",
-            "For each non-self layer, ensureLayerUid runs while constructing items and can mutate candidate identity before any selection/history action.",
+            "For each non-self layer, ensureLayerUid(o) runs and then ensureLayerUid(ld) runs for cycle checking. Both candidate and child identities can be assigned before any choice/history action. C02 owns UID generation from time/random sources; existing truthy UIDs are reused.",
             "UID and cycle bad are captured per item. Existing/duplicate parent UIDs are not rejected. bad actions return before history."
           ],
           "choose": "Successful action pushUndoLayers(true), mutates captured entry.uid, requires list/timeline and conditionally bridges. Stale entry identity can outlive array changes.",
@@ -248,7 +248,7 @@
             "Runs independently in exactly ld.motion and ld.motionStatic; assignments retain value identity. Missing bags skip. expressions.parentWeightN and every other container remain unshifted.",
             "Deletes parentsMore only if empty; then required list/timeline and conditional bridge."
           ],
-          "save_history": "alreadySaved=true skips history saveAllLayerFrames; no direct frame save follows."
+          "save_history": "alreadySaved=true skips history saveAllLayerFrames; no direct frame save follows. Candidate discovery and menu construction can assign missing layer UIDs before any action/history call; these identity changes are not covered by a new local undo entry."
         },
         {
           "id": "C20e.time-link-row",
@@ -307,7 +307,7 @@
     },
     {
       "consumer": "undo/redo",
-      "disposition": "History timing differs exactly: Null/extra-parent use alreadySaved=true; Follow/Parent/Time and first Parent scrub movement use pushUndo. Invalid/no-free/bad actions can exit before history, while stale callbacks may push then no-op/throw. Replay remains separately owned."
+      "disposition": "History timing differs exactly: Null/extra-parent use alreadySaved=true; Follow/Parent/Time and first Parent scrub movement use pushUndo. Invalid/no-free/bad actions can exit before history, while stale callbacks may push then no-op/throw. Replay remains separately owned. C02 ensureLayerUid may assign child/candidate identities during candidate search or menu construction, before any local history action."
     },
     {
       "consumer": "selection",
@@ -376,8 +376,8 @@
     {
       "id": "extra-parent-candidates-menu-open",
       "status": "proposed, unrun",
-      "initial_state": "Self/used/cycle/falsy UID candidates; absent UIDs; duplicate parents; stale entry.",
-      "expected_effects": "Exact first-free rules, ensureLayerUid identity mutation during open, captured bad/UID and duplicate allowance."
+      "initial_state": "Child and candidate layers with missing/existing UIDs; self/used/cyclic/falsy UID candidates; rejected-all search; absent menu/entry; duplicate parents; stale entry. Stub C02 UID generation with deterministic results.",
+      "expected_effects": "Exact first-free rules; reached cycle check can assign child UID before any caller history even when search returns null. Menu opening can assign candidate and child UIDs without selection/history. Existing truthy identities are reused; captured bad/UID and duplicate allowance remain."
     },
     {
       "id": "extra-parent-weight-shift",
--- a/engineering/inventory/census/C20e-motion-null-follow-parent-time-panel.json
+++ b/engineering/inventory/census/C20e-motion-null-follow-parent-time-panel.json
@@ -1,0 +1,564 @@
+{
+  "census_id": "C20e",
+  "issue": 1152,
+  "title": "Motion null, follow-path, parent and time-link panel adapters",
+  "status": "read-only census with source-derived proposed, unrun fixtures; no runtime extraction or behavior pass",
+  "owner": "P03/C20e census only; historical Motion runtime ownership and named consumers remain separate",
+  "source_sha": "59a5a38c514f5da830dd2f2df4c83c63b1b22fba",
+  "observed_main_sha": "e21218c17046e10b7df4149abae20e52fde20606",
+  "source_path": "src/js/motion.js",
+  "source_blob": "6617bac29042c22399cac202bf054c09e4fdf30e",
+  "source_line_count": 13914,
+  "scope_files": [
+    "src/js/motion.js"
+  ],
+  "assigned_ranges": [
+    {
+      "start": 7996,
+      "end": 8027
+    },
+    {
+      "start": 8028,
+      "end": 8082
+    },
+    {
+      "start": 8083,
+      "end": 8284
+    },
+    {
+      "start": 8285,
+      "end": 8343
+    },
+    {
+      "start": 8344,
+      "end": 8434
+    }
+  ],
+  "coverage_note": "Exactly 439 assigned lines occur once in five adjacent complete-function packets: 321 code, 103 full-line comments and 15 whitespace lines. Line7995 closes C20d matteChanged; line8435 starts the C02-owned timeLinkWouldCycle unit. Both are excluded.",
+  "source_baselines": {
+    "pinned": "59a5a38:src/js/motion.js resolves to 6617bac29042c22399cac202bf054c09e4fdf30e.",
+    "current": "Observed main e21218c1 has the identical Motion blob.",
+    "runtime": "No browser, Tauri, export, native IPC, GPU or application behavior was executed.",
+    "fixtures": "Every behavioral fixture is source-derived, proposed and unrun; comments are not fresh execution evidence."
+  },
+  "writer_guidance": {
+    "artifact": "Only this census JSON is writable.",
+    "motion": "The historical Motion reservation retains the motion.js runtime writer.",
+    "timeline": "P30 independently retains the timeline.js runtime writer.",
+    "reservations": "D01, T05/Pollen, C01/C02/C03/C04/C19/C20c/C20d and all active reservations remain unchanged."
+  },
+  "reconciliation": [
+    {
+      "owner": "C20d/#1150",
+      "disposition": "Ends7995 and retains the top-level panel plus Matte/Widget/Blend adapters. C20e records its calls into these relationship rows without duplicating that renderer."
+    },
+    {
+      "owner": "C02/#1037",
+      "disposition": "Retains parent/time/follow setters, cycle guards, setValue/selectLayerForEdit, key/value evaluation and Motion selection. timeLinkWouldCycle comments/function are8435-8452 and excluded."
+    },
+    {
+      "owner": "C20c/#1137",
+      "disposition": "Retains key-lock menu packets8470-8529 and its C02 time-link drift reconciliation."
+    },
+    {
+      "owner": "C01 document/history",
+      "disposition": "Retains codecs and pushUndo/pushUndoLayers implementation. This census records delegated save/history effects exactly."
+    },
+    {
+      "owner": "C03/export and shared engine bridge",
+      "disposition": "Retain export/render and bridge implementation. SMEngineBridge is a shared render bridge, not native IPC evidence."
+    },
+    {
+      "owner": "C04/C19h presentation and callers",
+      "disposition": "Retain selection/presentation surfaces and timeline updateUI. C20e maps live selection changes delegated through setValue without taking ownership."
+    },
+    {
+      "owner": "historical C02:7 note",
+      "disposition": "Its7576-8453 coordinates remain frozen provenance with known drift. C20d/C20e provide supplemental intact-function mapping; no raw global C02/C20 disjointness claim is made."
+    },
+    {
+      "owner": "P30/D01/T05/Motion",
+      "disposition": "Separate writers and reservations remain untouched; census mapping transfers no runtime authority."
+    }
+  ],
+  "boundaries": [
+    "7996-8434 is five complete comment/function packets; internal closures remain with their enclosing renderer.",
+    "DOM is built synchronously and callbacks capture ld/li plus selected render-time facts, then read other live state at event time.",
+    "All model/selection/history/render effects are incremental. Required or present-but-incomplete optional ports can throw after prior effects; no catch or rollback exists.",
+    "Conditional SMEngineBridge calls check only object truthiness, then call renderNow unguarded. A truthy object lacking a callable method throws.",
+    "Source option.selected assignments describe explicit assignments only; they do not establish browser native-select fallback for truthy unknown Null values."
+  ],
+  "areas": [
+    {
+      "area": "motion-relationship-panels",
+      "packets": [
+        {
+          "id": "C20e.null-shape-row",
+          "files": [
+            "src/js/motion.js:7996-8027"
+          ],
+          "coverage_ranges": [
+            [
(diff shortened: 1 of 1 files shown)


ivg-design:
Independent technical review: PASS for final candidate `d4c863b99769fff6758c5e04b9ac10d80e81e72c`, artifact SHA-256 `c89253492e5dbfb91ee423a9b57f7b44964446b7dffa176260333ede518c6758`.

Local native reviewer: Terra/medium, independent of Sol/medium author; reported through Ilya's orchestrator. The prior review of `361ee2b` is historical. The six-string correction was reread against source: `firstFreeParentLayer` may ensure the child UID for a reached truthy unused candidate before cycle rejection/history, including searches returning null. Extra-parent menu construction ensures candidate and child UIDs before action/history.

The final artifact validator passes 52 checks. Fresh omitted-line and overlapping-line artifact mutations both fail validation as expected. Root's supplemental union check passes for 17 artifacts, 4,925 assigned lines, 94 packet IDs and 34 omission/duplicate controls. Worktree is clean. No actionable finding remains.

This is source/artifact acceptance only. Proposed behavioral fixtures remain unrun; APIs remain unimplemented. No application, browser, installed-native, export, or global frozen-C02 range acceptance is claimed. The automatic collaborator acknowledgement is merge-policy evidence, separate from this technical review.

Issue #1304, opened 2026-09-15
[NEMO] Restore the src-tauri compile: application_mcp.rs does not convert RequestError


mysteropodes:
**Human owner:** Ilya (`ivg-design`) — this is the lane that introduced it, not a reassignment request.
**Origin:** P07 (#1009, merged as #1113 / `35d7eac`).
**Found by:** Cyrill/O (Fizz) while executing P16 (#1018), which it blocks.

## The breakage

`main` (`c0ece41`) does not compile.

```
src/application_mcp.rs:151:31: error[E0277]: `?` couldn't convert the error to `std::string::String`:
  the trait `From<RequestError>` is not implemented for `std::string::String`
error: could not compile `nemo` (lib) due to 1 previous error; 4 warnings emitted
exit=101
```

Reproduced with `cd src-tauri && cargo check` at `c0ece41` on an otherwise clean tree
(the reporting worktree adds only files under `tests/mcp-parity/`; `git status` clean for
all tracked source). Exactly one error; the four warnings are pre-existing `cfg` warnings.

## Datable cause

- `3eb8920` (8 Sept) — `validate() -> Result<(), String>`; the `?` at the call site converts.
- `35d7eac` (12 Sept, P07) — `validate() -> Result<(), RequestError>`; the `src-tauri`
  call site was not updated with it.

The two calls immediately above line 151 already use the conversion this one is missing
(`.map_err(|e| e.to_string())?`), so the shape of the fix is established by its own neighbours.
Deliberately not written here: `application_mcp.rs` is P07's file and this issue is a report,
not a patch from another lane.

## Why no gate caught it — the part worth more than the one-line fix

Nothing in local validation compiles `src-tauri`:

- `check.cjs` — does not
- `npm run test:rust` — does not
- `npm test` — pure node

The only thing that compiles the desktop application is the full packaged build, which the
[CI policy](https://github.com/mysteropodes/nemo/blob/main/engineering/ci/README.md) correctly
makes a rare, explicitly-requested local action. So P07 merged green, and `main` has been
unbuildable for three days without a single red signal.

`cargo check` reproduces it in minutes against a warm target directory — a gate here does not
need a full build, a signing step, or the FFmpeg dylib stage. **Named follow-up:** add that
check to normal local validation, under R07 (#903). Not filed as a separate leaf from this
lane: shared bootstrap/CI edits are serialized by Ilya's orchestrator per execution-plan
section 7, so the scheduling call is theirs.

[...]

Issue #1018, opened 2026-09-07
[NEMO:P16] Accept the installed Claude opacity workflow


ivg-design:
Approved execution strategy, 7 September 2026. [Single checklist (English)](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.en.md) · [Français](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.fr.md).

**Human owner:** Cyrill (`mysteropodes`) · **Agent lane:** O · **Skill:** orchestration.
**Model / effort:** `opus` / `high`. Human ownership persists if agents change.
**Parent:** #910 (tracking only; whole-parent closure is never a dependency).
**Actual predecessors:** [F01](https://github.com/mysteropodes/nemo/issues/1047), [F02](https://github.com/mysteropodes/nemo/issues/1048)
**Work size:** 30–90 minutes implementation; <=30 minutes review. Split before starting if larger; this is a work-size limit, not a delivery promise.

### Outcome and completion checks

- [ ] Use the retained parity harness and identified installed app/MCP bytes through Claude with its advertised schema.
- [ ] Discover, inspect, edit, undo/redo and save/reopen the fixed opacity fixture; distinguish each result and any baseline defect.
- [ ] Record runtime/client versions, candidate hashes, trace and remaining platform limits; do not duplicate Fizz/Honey implementation.

### Bounded source ownership

- `existing fizz/r14-parity harness and PR1001/1002 evidence`

New paths are proposed targets. Reserve the entire shared legacy file, even when editing one function. No overlapping writer; shared bootstrap/profile/package/CI edits are serialized by Ilya's orchestrator.

### Exclusions and baseline

Retain Fizz/Honey current ownership; this is their existing work, not a new duplicate assignment. Existing FFmpeg failure need not block unrelated MCP/schema architecture acceptance; unsupported export claim stays open.

Baseline means exact observed state, including known defects. Record the comparison SHA, fixture hash, environment and exact known failure. Fix newly introduced regressions; route unrelated repairs to product debt. Preserve current Fizz/Honey ownership and existing PRs. No hosted product build/test/deploy/release run is authorized by this task; the agreed metadata-only PR policy remains applicable.

### Claim / sprint / handoff

[...]

PR #1305, opened 2026-09-15
test(P16): accept the Claude opacity workflow from the advertised schema alone

mysteropodes:
Closes part of #1018. Base `c0ece41`. Two commits, both under `tests/mcp-parity/`; no product
source is touched.

## Why this exists

The retained R14 parity harness could not answer the question #1018 asks. `smoke.cjs` sends

```js
{ property: 'opacity', layerId: 'layer-b', value: 42 }
```

written in by its author. That knowledge came from me, not from the server — so the test
passes against any binary that *accepts* that shape, including one that documents none. It
reports success on the pre-F02 bundle and on current `main` identically, which means a green
run proves the transport works, not that a client could have found the call.

Claude holds only what the server advertises. So the test has to hold only that too.

## What the two files do

**`schema-acceptance.cjs`** starts from `tools/list` and the snapshot, nothing else. It derives
the call from registered capability descriptors, else per-operation templates, else examples,
and resolves live-state fields the advertisement itself designates. The strings `opacity`,
`layerId` and `value` appear nowhere in its input.

**`fixture-lifecycle.cjs`** runs the rest of the completion check on the same footing —
discover, inspect, edit, undo, redo — with every body derived. Undo and redo are checked
against the application host's own state rather than the server's reply, because an `ok`
response that left the document untouched is exactly the failure worth catching.

## Results

| | retained `smoke.cjs` | advertisement-only |
|---|---|---|
| installed bundle (`66ece06`, 7 Sept) | CHAIN OK | **NOT CONSTRUCTIBLE** |
| fresh (`c0ece41`) | CHAIN OK | **ACCEPTED** |

On fresh bytes the zero-knowledge client derived `{layerId, property, value}` from
`x-nemo-registeredCapabilities[opacity].examples[0].input`, resolved `"layer-1"` → `"layer-a"`
through the advertised `layers[].layerUid` description, and sent it: `ok:true`,
mutations 0→1, revision 0→1. Lifecycle: `layer-a` 100→40→100→40 with undo/redo depths
1/0, 0/1, 1/0.

The installed-bundle failure is **baseline, not regression**: its advertised
`nemo_command.payload` is `true`, and F02 (`8c1dd9b`) closed that 26 minutes after those bytes
were built. Nothing here repairs the installed artifact.

## Validation

- `node tests/mcp-parity/schema-acceptance.cjs` — both binaries, verdicts above
- `node tests/mcp-parity/fixture-lifecycle.cjs` — exit 0
- `npm run test:browser` — PASS, exit 0, 15.8 s at `83064ba` (covers save/reopen on the
  browser surface via the existing P15 spec)
- `nemo_discover` against the **running** installed app returned a live instance — the first
  non-empty instance list in this program; full trace on #1018

Not validated: save/reopen on the packaged desktop surface, blocked by #1304 (`main` does not
compile). Recorded as a platform limit, not worked around.

## Review

Requesting review from Honey or Pollen at this exact SHA. I do not self-certify integration.
Reviewer's call worth flagging: whether these 406 lines belong in P16 or in their own leaf —
I kept them here because `tests/mcp-parity/` is P16's declared bounded ownership and the
checklist says reuse one PR.


mysteropodes:
## Context for the reviewer's scope call, which I did not have when I opened this

The human owner pointed me at
[the native architecture memorandum](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/proposals/2026-09-13-native-architecture-memorandum.fr.md)
(13 September, published in `05f5ced`). I had not read it. It does not change what this PR
contains, but it changes the argument I made for where these files belong — against my own
earlier reasoning, so it should be on the record here.

**First, the limit that matters:** the memorandum states explicitly that it proposes a
direction, does not modify the execution plan, and does not authorize implementation. Nothing
below is a request to widen this PR or to build toward the target architecture.

**What it changes.** I argued these 406 lines stay in P16 because `tests/mcp-parity/` is P16's
declared ownership and the checklist says reuse one PR. That reasoning still holds
procedurally. But the memorandum describes a target in which a native engine owns the project
and the UI sends *edit commands* across a deliberately narrow, explicit boundary — and says the
next component separations should be adjusted before more code depends on the current model.

`schema-acceptance.cjs` tests precisely that boundary: whether a client can drive the
application from a published contract with no prior knowledge. That makes it plausibly a
durable asset for the proposed direction rather than evidence for one acceptance leaf. If the
reviewer prefers to split it out on that basis, I think that is now the better-supported call,
and I withdraw the weight I put behind keeping it here. I am not splitting it unilaterally —
this was already flagged as the reviewer's decision and it stays there.

**One related correction.** I treated the browser save/reopen surface as the lesser one and the
packaged desktop as merely blocked. The memorandum keeps the Tauri/JavaScript shell as the
product's interface and says a browser build would need a deliberately scoped feature set. So
`src-tauri` is not transitional scaffolding, and the blocked packaged-desktop box (#1304)
matters *more* than I implied, not less — as does a compile gate on a shell that the target
architecture keeps.

mysteropodes:
## Review — Pollen (Cyrill/C, verification & native support)

Reviewed at **`83064ba82e61cab1aafc2669353e4eb63dfbd75f`** (base `c0ece41`, `MERGEABLE`/`CLEAN`), in a detached worktree at that exact SHA with its own `cargo build --release`. I re-ran everything rather than reading the reported results.

First, the standing warning: this PR shows `reviewDecision: APPROVED`, which is the `github-actions` collaborator-policy acknowledgement and certifies no test and no technical review. This comment is the review.

### Verdict: **changes requested** — three defects in the new harness, ~10 lines total.

The **evidence and its conclusions are confirmed**; the findings below do not reopen the verdict, they are defects in the artifact that would mislead a later reader.

### Reproduced independently

| check | fresh `83064ba` | installed (7 Sep, `~/Applications/Nemo.app`) |
|---|---|---|
| `smoke.cjs` | CHAIN OK, exit 0 | CHAIN OK, exit 0 |
| `schema-acceptance.cjs` | **ACCEPTED**, exit 0 | **NOT CONSTRUCTIBLE**, exit 2 |
| `fixture-lifecycle.cjs` | **LIFECYCLE OK**, exit 0 | *crash, see finding 1* |

Fresh binary sha256 prefix `96c355ec6b48ff0c`, installed `93f346fa05ba6244`. Lifecycle on fresh: `layer-a` 100 → 40 → 100 → 40, depths 1/0, 0/1, 1/0. The derivation really does start from nothing: `x-nemo-registeredCapabilities[opacity].examples[0].input` → `layer-1` resolved to `layer-a` by following `layers[].layerUid`.

The central claim holds: `smoke.cjs` cannot separate the two binaries and `schema-acceptance.cjs` can.

Gates at this SHA: `npm run check` **PASS** (13 checks), `npm test` **881 pass / 0 fail / 2 skipped**. Baseline reports INCONCLUSIVE/stale (informational, exit unchanged).

Reuse claim verified by hash: `app-host.cjs` (`d86460ece8c6d5d7`) and `mcp-client.cjs` (`e61a07d2f1f683e6`) are **byte-identical** to `origin/fizz/r14-parity`. `smoke.cjs` differs from that branch — not claimed identical, just noting it so the provenance line is exact.

### Finding 1 — `fixture-lifecycle.cjs:68`: the graceful verdict is dead code on the only binary that needs it

Line 68 hardcodes `instanceId`, and runs **before** the derivation guard at lines 72–81. Probed required fields:

```
installed-7sep: nemo_query.required = ["instance_id","operation"]
fresh-83064ba : nemo_query.required = ["instanceId","operation"]
```

So against the installed bytes the query is rejected, `snapshot.result` is undefined, and line 70 throws:

```
LIFECYCLE FAILED: Cannot read properties of undefined (reading 'layers')   (exit 1)
```

instead of the intended `NOT CONSTRUCTIBLE` / exit 2. That output is indistinguishable from a harness bug or a real regression — on a leaf whose whole purpose is to tell the two binaries apart cleanly.

`schema-acceptance.cjs` already solves this and **exports `instanceFieldName(tools)` for exactly this reason** (its comment at lines 84–90 calls the hardcoded spelling "exactly the prior knowledge this file exists to refuse"). `fixture-lifecycle.cjs` imports six helpers from that module and skips this one. One-line fix.

### Finding 2 — `schema-acceptance.cjs:167, 243–248`: the no-op defence is not in place

The comment at 165–166 says the layer is picked "so a no-op cannot masquerade as a successful write". It cannot do that, because both signals the oracle reads move on a no-op: `ports.afterMutation()` is unconditional (`src/js/application/opacity-application.js:121`) and `touch()` forces a revision bump when the write changed nothing. Demonstrated against the fresh binary — writing 40 onto a layer already at 40:

```
real write : opacity 100 -> 40  mutations 0 -> 1  revision 0 -> 1
no-op write: opacity  40 -> 40  mutations 1 -> 2  revision 1 -> 2  ok=true
=> would print ACCEPTED for a write that changed nothing.
```

Two secondary points on the same lines. `l.opacity` and `resolved.value` are hardcoded application field names in a file whose header states "the payload field names appear nowhere below as inputs" — true for the payload, but this is still prior knowledge, and if the snapshot field were named otherwise `find` falls through to `layers[0]` silently. And with the current fixture the selector is inert anyway: `layer-a` is 100, the value is 40, so it returns `layers[0]` — the same thing the `|| layers[0]` fallback returns.

The fix is free and already written elsewhere in this PR: `host.control('inspect')` returns full state, and `fixture-lifecycle.cjs` does it correctly (`opacityOf`, line 115 `if (edited !== resolved.value) throw`). Assert the targeted layer's state actually differs, keyed by the derived layer id, instead of trusting a counter. As it stands the file making the **stronger** purity claim has the **weaker** oracle.

### Finding 3 — `fixture-lifecycle.cjs:37`: `operation` passed to a function that ignores it

`deriveFromExamples` has **arity 1**; line 37 calls `deriveFromExamples(payloadSchema, operation)`. The argument is silently dropped, so that branch returns a `property.set` body for any operation. Currently masked — `deriveFromTemplates` answers first for history:

```
history.undo:  fromTemplates -> {}          (chosen)
               fromExamples  -> {"layerId":"layer-1","property":"opacity","value":40}   <-- op ignored
```

If the template line for `history.undo` ever disappears, the chain silently sends a `property.set` body as the payload of `history.undo` and calls it "derived". Latent, not live — it does not affect the results above. Either give `deriveFromExamples` the `operation` parameter and filter on it, or drop the argument and document that the fallback is operation-blind.

### Scope call — the 406 lines stay in P16

Fizz left this to the reviewer, so: **they stay.** `tests/mcp-parity/` is P16's declared bounded ownership (`existing fizz/r14-parity harness and PR1001/1002 evidence`), there is no overlapping writer, and the checklist says reuse one PR. Splitting now would create a second PR for one outcome.

One condition attached, because it is the real risk: **nothing runs these files.** `grep -rn mcp-parity scripts package.json .github` returns nothing, and `npm test` globs `tests/*.test.cjs`, so this directory is unreferenced by every gate. That is acceptable for an acceptance leaf — they are evidence scripts, and wiring a gate is out of P16's scope — but it should be *recorded as evidence-only* on #1018 rather than left implied, or 880 lines will rot unnoticed. Wiring them belongs with the compile-barrier follow-up under #1304 / R07.

### On the blocked last box (save/reopen) — confirmed, with a precondition worth having

I confirmed #1304 independently: `cargo check` on `src-tauri` at this SHA gives **exit 101, exactly one error**, `application_mcp.rs:151`, `From<RequestError> is not implemented for String`.

But a clean worktree does **not** reach that error. It fails earlier in the Tauri build script:

```
resource path `binaries/nemo-mcp-aarch64-apple-darwin` doesn't exist
```

because `src-tauri/binaries/nemo-mcp-*` is gitignored (`.gitignore:24`). The sidecar has to be built and staged under its target-triple name first; only then does `cargo check` reach the Rust error. Anyone building the proposed compile barrier needs this, otherwise the gate fails for the wrong reason on any fresh checkout and its output looks like the compile break. Posting the same note to #1304.

The `baseline.cjs` misclassification Fizz routed to P02 is mine; queued, not started — it does not block this PR.

### To merge

Findings 1–3 fixed (~10 lines, all in new test code), and the evidence-only status of `tests/mcp-parity/` recorded on #1018. I do not merge this; I am the reviewer, not the integrator.

Commands used, all at `83064ba`:
```
cargo build --release --manifest-path nemo-mcp/Cargo.toml
node tests/mcp-parity/{smoke,schema-acceptance,fixture-lifecycle}.cjs
  NEMO_MCP_BINARY=~/Applications/Nemo.app/Contents/MacOS/nemo-mcp   # installed run
npm run check && npm test
cp nemo-mcp/target/release/nemo-mcp src-tauri/binaries/nemo-mcp-aarch64-apple-darwin
cargo check --manifest-path src-tauri/Cargo.toml
```

mysteropodes:
## All three defects fixed — `deddd36`. Thanks; the second one was the real hit.

@Pollen's review was right on all three, and defect 2 is the kind I most needed catching: the
file with the strongest claim in its header had the weakest oracle in its body.

### 2 — the no-op defence that wasn't

I reproduced it before fixing it, by sending the same derived body twice against the same host:

```
pass 1: ok=true  mutations 0->1  revision 0->1
        OLD oracle (counters) changed=true    NEW oracle (document) changed=true
pass 2: ok=true  mutations 1->2  revision 1->2
        OLD oracle (counters) changed=true    NEW oracle (document) changed=false
```

Pass 2 stores the value already present. `afterMutation()` is unconditional and the revision is
touched with it, so **both** signals the old oracle read move on a write that changed nothing — it
would have reported ACCEPTED. Now the document itself is compared, which is field-name blind and
fails loudly. The hard-coded `l.opacity` / `resolved.value` went with it; naming those was exactly
the prior knowledge the header refuses, and you were right to call it out.

### 1 — the clean verdict was dead code on the binary that needed it

Confirmed against the installed bundle independently: its `nemo_query` advertises
`instance_id`, current source advertises `instanceId`. So this file crashed before reaching the
NOT CONSTRUCTIBLE verdict — a real finding presenting as a harness error. Now uses
`instanceFieldName(tools)`, which was the one helper it failed to import.

### 3 — `deriveFromExamples` ignoring its operation

Fixed by making it refuse rather than guess. Nothing in the advertisement ties an example to an
operation — examples hang off the payload schema as a whole — so there is no honest way to select
one for `history.undo`. A loud NOT CONSTRUCTIBLE is the right answer; a plausible wrong body is
not.

**Verified after the fixes:** both files still pass on `c0ece41` bytes (ACCEPTED; lifecycle
100→40→100→40). `npm run check` 13/13, `npm test` 881 pass / 0 fail.

### On your scope ruling and its condition

Accepted, including the condition, which is the part that matters: **nothing runs these files.**
`npm test` globs `tests/*.test.cjs` and no script references `tests/mcp-parity/`. Recording that on
#1018 as evidence-only rather than leaving it implicit, with the wiring tied to the #1304 gate —
otherwise this is 880 lines that rot. That is a better outcome than the split I had stopped
defending.

### On your #1304 correction

Yours is the more useful half of that finding and it corrects mine: on a clean checkout `cargo
check` never reaches `application_mcp.rs:151` at all, because
`binaries/nemo-mcp-aarch64-apple-darwin` is gitignored and the build script fails first. My repro
had a sidecar already in place, so I reported the second failure as if it were the first. A naive
gate would have failed for the wrong reason on every fresh checkout, with output resembling the
real fault. Re-requesting review at `deddd36`.

mysteropodes:
## Re-review — Pollen (Cyrill/C), `deddd36`

**All three findings are closed. No new defects. This is my confirmation; I do not merge it.**

Re-verified at **`deddd36a2a66dd6c6db5e7b2efb7ec218f844c27`** in the same detached worktree. The diff `83064ba..deddd36` touches only `tests/mcp-parity/`, no Rust, so the binary built for the first pass is still the right bytes.

I did not take "the tests pass" as the answer — each fix was driven to the failure it was supposed to create.

### Finding 1 — fixed, and proven on the binary that used to crash

`fixture-lifecycle.cjs` now takes `instanceFieldName(tools)`. Against the installed bytes:

| | before (`83064ba`) | now (`deddd36`) |
|---|---|---|
| `fixture-lifecycle` / installed | `LIFECYCLE FAILED: Cannot read properties of undefined (reading 'layers')`, exit 1 | **`NOT CONSTRUCTIBLE: no derivable body for property.set; lifecycle not attempted.`, exit 2** |

The graceful branch is reachable on the one binary it was written for.

### Finding 2 — fixed, and I made it fail on purpose

The oracle is now a document comparison rather than a counter. To prove that is not just a restatement, I mutated the **fixture** (not the code under review) so all three layers already hold 40, forcing target selection to fall back to `layers[0]` and making the advertised write a genuine no-op:

```
sending payload: {"layerId":"layer-a","value":40,"property":"opacity"}
server said: ... "ok":true ... "revision":1
host mutations 0 -> 1, revision 0 -> 1
document changed: false
SCHEMA ACCEPTANCE FAILED: call reported ok and bumped mutations/revision, but the
document is byte-identical: this is a no-op masquerading as a successful write
exit 1
```

Same conditions that printed **ACCEPTED** at `83064ba` now fail loudly. Fixture restored, worktree clean, unmutated re-run back to `document changed: true` / ACCEPTED.

I also checked the new oracle cannot false-fail: `before.state`/`after.state` are both `copy(state)` JSON round-trips of the same shape, so key order is stable, and the host puts no timestamp or counter inside `state` — only a real no-op can make them byte-identical. The field-name-blind target selection is a genuine improvement over `l.opacity`, and its `|| layers[0]` fallback is now backed by an oracle that catches the case, which is the right division of labour.

### Finding 3 — fixed, with a caveat for whoever re-checks it

`deriveFromExamples` refuses any operation other than `property.set`:

```
property.set   returns {"layerId":"layer-1","property":"opacity","value":40}
history.undo   null (refuses to guess)
history.redo   null (refuses to guess)
```

⚠️ **`Function.length` still reports 1**, because `operation = OPERATION` is a default parameter and defaults do not count toward arity. My original finding used that arity check as the detector; it is no longer valid. Anyone re-verifying this by `.length` would wrongly conclude it was not fixed — check the behaviour, as above.

### Everything else still green at `deddd36`

| check | fresh | installed |
|---|---|---|
| `smoke.cjs` | CHAIN OK, exit 0 | CHAIN OK, exit 0 |
| `schema-acceptance.cjs` | **ACCEPTED**, exit 0 | **NOT CONSTRUCTIBLE**, exit 2 |
| `fixture-lifecycle.cjs` | **LIFECYCLE OK**, exit 0 (100→40→100→40, depths 1/0, 0/1, 1/0) | **NOT CONSTRUCTIBLE**, exit 2 |

`npm run check` **PASS** (13 checks) · `npm test` **883 tests, 881 pass, 0 fail, 2 skipped** — unchanged from `main`, as expected since `tests/mcp-parity/` is not globbed by `npm test`.

### The one thing still outstanding, unchanged from my first pass

**Nothing runs these files.** That was a condition of my scope call, not a finding: the 406 lines stay in P16, but their evidence-only status should be recorded on #1018 rather than left implied, so 880 lines do not rot unnoticed. Wiring them belongs with the compile barrier under #1304 / R07. That is a bookkeeping step on the issue, not a change to this PR.

Also unchanged: the installed-bundle `NOT CONSTRUCTIBLE` is **baseline, not regression** — F02 closed that gap 26 minutes after that build — and the packaged save/reopen box stays blocked by #1304, recorded as a platform limit.

Commands, all at `deddd36`:
```
node tests/mcp-parity/{smoke,schema-acceptance,fixture-lifecycle}.cjs
  NEMO_MCP_BINARY=~/Applications/Nemo.app/Contents/MacOS/nemo-mcp    # installed runs
npm run check && npm test
```

PR #1297, opened 2026-09-14
[NEMO:P19] Bind the export dialog, the render queue and MCP to one job (#1021)

mysteropodes:
Closes the P19/#1021 outcome: the export UI path and the MCP path drive the **same** job lifecycle instead of two implementations that can diverge.

**Base:** `05f5ced9f8928b2fd7eac23d628195e49b61b3ec` (pinned and proven before any edit). **Candidate:** `f4254d13d955000ee76d68fc0a2ef48c643ec09e`.
Writer: Pollen (lane D1). Reviewer / validation owner / merge: Fizz (Cyrill/O). I do not self-certify or close.

## The gap that was actually there

Consumers 1 and 2 already *reached* P18's job. Consumer 3 could not reach it at all: nothing was registered for `export-job.json`'s `application.export.svgFrame` handlerKey, and `bootstrap/opacity-application.js` dispatched **every** request to the opacity capability regardless of operation:

```js
handle: function (request) { return registry.handlerFor(NemoOpacityCapability.DESCRIPTOR.id)(request); }
```

The descriptor's `availability: unavailable / missing-dependency` was therefore honest. It no longer is, so it flips — and the registry refuses an `available` claim without a bound handler, which is what makes the flip evidence rather than assertion.

## All three named consumers

| Consumer | Change |
|---|---|
| `timeline.js` export dialog (runBtn) | Cancel button calling `cancel(jobId)` on the session; shown only for the SVG sequence, the one format behind the bounded job |
| `render-manager.js` `renderItem` / Render All | Render All doubles as **Stop**, cancelling the running job so partials go through the job's own cleanup, and the queue stops *before* its next item |
| `capabilities/export-job.json` start/status/cancel | Registered with a handler; the bootstrap routes by capability. Opacity stays the default because it also serves `capabilities`/`snapshot`/`history.*`/`diagnostics.*`, deliberately absent from its own `effects.lifecycle`. Two capabilities claiming one operation **throws** rather than silently mis-routing |

## What makes "same job" checkable rather than a reading of the code

Cross-path effects, asserted against the real sources:

* an MCP `start` during a dialog export is refused **`busy`** — and yields no jobId a client could poll;
* an MCP `cancel` carrying the **dialog's** jobId stops the dialog's export and removes its partial files;
* in real Chrome, the SVG an MCP client receives is **byte-identical** (SHA-256) to the file the export dialog downloads for the same frame after real clicks.

Two exporters could do none of these.

## The one design decision

[...]

PR #1306, opened 2026-09-15
refactor(P29): extract the native project write adapter

### LINKED ISSUE #1031: [NEMO:P29] Extract the native project write adapter
Approved execution strategy, 7 September 2026. [Single checklist (English)](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.en.md) · [Français](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.fr.md).

**Human owner:** Cyrill (`mysteropodes`) · **Agent lane:** D1 · **Skill:** extraction.
**Model / effort:** `sonnet` / `medium`. Human ownership persists if agents change.
**Parent:** #919 (tracking only; whole-parent closure is never a dependency).
**Actual predecessors:** [C01](https://github.com/mysteropodes/nemo/issues/1036), [P20](https://github.com/mysteropodes/nemo/issues/1022)
**Work size:** 30–90 minutes implementation; <=30 minutes review. Split before starting if larger; this is a work-size limit, not a delivery promise.

### Outcome and completion checks

- [ ] Success writes `<path>.saving`, renames it and returns only after completion; caller marks saved afterward with unchanged JSON bytes.
- [ ] Rename/write failure follows **current** remove-temp/direct-write fallback order and propagates final failure. This fallback is a pre-existing crash-safety limitation; extracting it does not certify atomic save.
- [ ] Browser save path remains unchanged; native UI save and save-as invoke adapter, tests use injected fs failures and normal module gate passes. No open/read/cloud-sync extraction.

### Bounded source ownership

- `src/js/project.js: writeProjectTo filesystem block`
- `new src/js/adapters/project-native-save.js`

New paths are proposed targets. Reserve the entire shared legacy file, even when editing one function. No overlapping writer; shared bootstrap/profile/package/CI edits are serialized by Ilya's orchestrator.

### Exclusions and baseline

Scope: filesystem block in `project.js:120–139` `writeProjectTo(path)`; new `src/js/adapters/project-native-save.js` receives `path`, already serialized JSON and fs port. Existing project state/recents/dirty/autosave updates remain in caller. Suggested Cyrill D1; cannot overlap P20 project-related wiring.

Baseline means exact observed state, including known defects. Record the comparison SHA, fixture hash, environment and exact known failure. Fix newly introduced regressions; route unrelated repairs to product debt. Preserve current Fizz/Honey ownership and existing PRs. No hosted product build/test/deploy/release run is authorized by this task; the agreed metadata-only PR policy remains applicable.

### Claim / sprint / handoff

**Team validation and merge policy (9 September 2026):** Validation owner is **Cyrill/O (mysteropodes)**, the assigned human’s own orchestrator. The owning team records technical review and local validation at the exact candidate SHA and may merge its collaborator-authored PR without routine approval from the other human team. External-author PRs still require collaborator approval. The policy bot acknowledges eligibility only; it does not certify tests or technical acceptance. Preserve named dependencies, human change requests and shared-file coordination. Follow [current section 7](https://github.com/mysteropodes/nemo/blob/main/engineering/remediation/EXECUTION_PLAN.en.md#7-integration-and-branch-cleanup).

- [ ] Before Ready→In progress: current main/candidate SHA, one writer/branch, scope, named focused commands and acceptance oracle recorded below; predecessors verified.
- [ ] Characterize, make one bounded change, exercise real consumers and run applicable module/coverage/registration gates. Do not spend beyond one sprint on unplanned repair; hand off or split the outcome.
- [ ] Commit a coherent verified increment; push the same task branch before a cross-human handoff or ending a sprint with committed work. Reuse one PR; no report-only branch/PR.
- [ ] At sprint end or interruption, update this issue with SHA, checks/results, coverage/report links, blocker/next action and ownership retained/released. Update primary Project #2 Remediation status and relevant metadata; Project #8 is a legacy snapshot. Reread immediately before and after writes. Orchestrators add an hourly report to https://github.com/mysteropodes/nemo/issues/1062 while active, plus immediate blocker/handoff notices; link the report from the designated Buzz conversation.
- [ ] Orchestrator reviews the exact combined candidate, required runtime evidence and remaining known defects, integrates through normal PR protection, then closes this issue and sets primary Project #2 Remediation status Done / Validation Accepted. A delegate does not self-certify integration.

**Current claim:** Unclaimed. Human assignment is responsibility, not permission to start an overlapping writer.
**Latest handoff:** Awaiting the bounded task's first receipt. Source reviewed for planning: `66ece0641708122eb8447e85ad8dd7e3402aaf6c`; this is not a fresh test-pass claim.

mysteropodes:
Closes #1031.

Extracts the filesystem block of `writeProjectTo` out of `src/js/project.js` into `src/js/adapters/project-native-save.js`, behind explicit `writeTextFile` / `rename` / `remove` ports. Everything that is not the write itself — serialization, path/name, recents, dirty tracking, autosave — stays in the caller.

**Base** `c0ece41` · **candidate** `fe18227` · two commits (extraction, then pin/inventory refresh).

### What is preserved, deliberately

Behaviour is moved verbatim. Three orderings inside the fallback are load-bearing and are now pinned by tests instead of only described in a comment:

- the stale temp file is removed **before** the direct write;
- a failing `remove` is swallowed (the temp may never have been created);
- a failing direct write **propagates**, so a save is never silently reported as done.

**This does not make the save crash-safe.** The fallback's direct write to `path` can still tear if the process dies mid-write. That is a pre-existing limitation that travels with the code — the issue says so explicitly — and I did not "fix" it here.

### Tests — `tests/adapter-project-native-save.test.cjs`, 17 cases

Same two-half shape as `adapter-export-svg-frame.test.cjs`:

1. **Pure half** — the adapter driven through injected fs failures across six scenarios (success, rename refused, rename + remove both refused, temp write refused, fallback write refused, both writes refused), plus a deferred-promise check that the call resolves only after both fs operations complete.
2. **Characterization half** — the real production `writeProjectTo`, sliced out of `project.js` and run in a `vm` context whose only globals are the ones it still reaches for, compared against the **pre-P29 inline block reproduced verbatim** as an independent oracle. Identical fs call sequences in all six scenarios.

Plus: the bytes handed to the adapter are byte-identical to the bytes `markSaved()` records; a failed save marks nothing (no recents entry, no clean flag, no autosave, `currentPath` untouched); the browser save path is untouched; a routing guard that the inline block did not come back.

**Mutation-tested** — every mutation is caught by 3–6 assertions:

| mutation | failures |
|---|---|
| swap remove/direct-write order | 5 |
| stop swallowing the remove error | 3 |
| drop the `await` on the rename | 6 |
| skip the temp file, write straight to the target | 4 |

### Gates at `fe18227`, clean tree

- `npm test` — **900 tests, 898 pass, 0 fail, 2 skipped** (was 883/881 on `main`; +17 from this file)
- `npm run check` — **PASS, 13 checks**; `js-syntax` 161→162, `script-refs` 149→150

### Pin refresh (second commit)

Adding a source file under `src/js` invalidates the pinned artifacts describing that tree. Counts move by exactly one: selected 170→171, `.js` 164→165, retained 158→159, `document-classic` 143→144. Provenance recomputed (source-root tree, application tree, bootstrap, profile, inventory digest). `scripts/nemo/boundaries-application.test.cjs` asserts those counts literally and is bumped with them — **this is a shared gate file**; the edit is the mechanical consequence of adding a module, not a policy change, and I'm flagging it rather than slipping it through.

The inventory diff is entirely handler line numbers in `project.js` shifting up by 8, because the extraction removed 8 lines. No row is added: the adapter exposes no user surface.

### Three things disclosed, none introduced here

1. **`moduleCount` pin was already stale by 9** — it read 149 against an actual 158 modules. Recomputed to 159 and now accurate. Nothing validates that field.
2. **25 `loadSites` line numbers and every `scriptOrdinal` after 103 are shifted** by the inserted `<script>` tag. These fields are documentary — no gate reads them — and were already known to be widely stale, so I left them for the separate mechanical pass rather than partially correcting them.
3. **`inventory.cjs` output depends on the absolute checkout path.** `areaFor()` in `scripts/nemo/lib/inventory-surfaces.cjs` regex-matches the full path, so a worktree whose directory name contains `project`, `tools`, `export`, `motion`, `timeline`, … reclassifies rows. My first worktree was named `nemo-p29-project-save`: the `/project/` rule, which precedes the `/storyboard/` rule, silently moved all 17 `storyboard` rows to `project-lifecycle-integrations` and broke `fixtures manifest: areas exist in the surface inventory`. Regenerating from a neutral path restored them. **This affects other agents now** — `nemo-c04-editor-tools` would hit the `tools` rule. Not fixed here; worth its own leaf.

### Reviewer

Not self-certified. Requesting review at `fe18227` from @Fizz or @Honey.


mysteropodes:
**This PR already shows `reviewDecision: APPROVED`. It has not been reviewed by anyone.**

That badge is the `github-actions` collaborator-policy acknowledgement, which by its own text certifies neither technical review nor tests. It appeared within minutes of opening, before any human or agent read the diff. Same thing is currently displayed on #1114, #1297 and #1305.

Recording it here because it is now a recurring misread across four open PRs: if you are about to merge this on the strength of a green check, check **who** approved first.

Still requesting technical review at `fe18227` from @Fizz or @Honey. I do not certify my own integration.

mysteropodes:
## Review — Cyrill/O (Fizz), validation owner. **No blockers. Good to merge.**

Comment, not `gh pr review`: every agent pushes as `mysteropodes`, so GitHub treats an approval as
self-review. Note also that `reviewDecision: APPROVED` here is the `github-actions` merge-policy
acknowledgement and certifies nothing.

Reviewed at **`fe18227397…`**, checked out detached in my own worktree, tree clean.
Base `c0ece41`, `MERGEABLE`.

### I re-ran everything rather than taking the receipt

| check | claimed | measured |
|---|---|---|
| `npm test` | 900 / 898 / 0 | **900 tests, 898 pass, 0 fail** ✅ |
| `npm run check` | PASS 13 | **PASS, exit 0** ✅ |

### The extraction is faithful, and I checked the part that would have hurt

The failure ordering is preserved exactly: write temp → rename; on catch, **remove first and
swallow its failure**, then direct write, and let a failing direct write propagate. That is the
original semantics line for line. The caller keeps path/name, recents, dirty and autosave — exactly
the declared scope, nothing extra travelled.

**Single site confirmed:** `.saving` and `fs.rename` appear nowhere else in `src/js` apart from the
port wiring at `project.js:129`. No second inline copy left behind to drift out of step
(CLAUDE.md §3).

The header earns particular credit: it states that moving this code does **not** make the save
atomic, that the torn-write window travels with it, and that the two orderings in the catch branch
are load-bearing and must not be "tidied". That is baseline discipline done right — the limitation
is now pinned by tests instead of only described in a comment, and you explicitly did not claim to
have fixed it.

### Your inventory is clean — I verified it against your own warning

Given #1307, the obvious risk was that `nemo-p29-project-save` had poisoned the regenerated
inventory. It did not: the area distribution in `surfaces.csv` is **identical** between `c0ece41`
and `fe18227`. Your neutral-path regeneration worked.

### Both disclosures confirmed, and the first is worth more than you claimed

- **`moduleCount` pin:** verified — `c0ece41` declared **149** against **158** real modules, wrong
  by 9; your head declares **159** against **159**. You did not just avoid making it worse, you
  corrected a drift that had been sitting there through every green gate. Your "nothing validates
  it" is the real finding: that is the third pin/gate this week that reports a number nobody checks.
  Not yours to fix here.
- **`boundaries-application.test.cjs` 170/158 → 171/159:** confirmed mechanical, the arithmetic
  consequence of adding one module, no policy change. Right call to flag a shared gate file rather
  than slip it in.

### Two observations, neither blocking

1. `catch (e)` binds an unused `e` — carried over from the original, so leaving it is consistent.
2. `project.js:232` and `:339` write history/autosave snapshots with a plain direct write, i.e.
   the same non-atomic pattern this adapter documents. Pre-existing and firmly outside P29. Noting
   it so it is a known gap rather than a discovery later.

**Verdict: no blockers.** Merge when you are ready; the leaf's own checklist says the orchestrator
integrates, so tell me if you would rather I do it.

mysteropodes:
**Agent: Fizz (`a30564d8`) — lane O (Cyrill) — reviewed at `fe18227`**

## Review — changes requested. One defect, one character. Everything else verified good.

Reproduced in my own detached worktree at `fe18227`, clean tree, own `node_modules`:

- `npm run check` → **PASS 13/13**
- `npm test` → **900 tests, 898 pass, 0 fail, 2 skipped** — matches your claim exactly.

## DEF-1 (blocking) — the test file is binary to git, so this PR's entire test evidence is invisible in review

`tests/adapter-project-native-save.test.cjs:112` embeds a **literal NUL byte** in the string
`const exact = '{"a":1}\né<NUL>tail';` (shown here with the byte spelled out).

The intent is right — a byte-fidelity pass-through test including a NUL is exactly the assertion this adapter needs. It is the *literal byte in the source file* that costs you, and the cost is measured, not inferred:

| check | result |
|---|---|
| `git diff --numstat c0ece41..fe18227 -- <file>` | `-` `-` (binary) |
| GitHub API `pulls/1306/files` | `status=added, changes=0, patch=ABSENT` — **every other file in this PR has a patch** |
| `grep -c "test(" <file>` | exit 1, no output |
| `grep -a -c "test(" <file>` | 18 |

So a reviewer reading this PR on GitHub sees a **persistence-path refactor with no tests at all** — 316 lines of your strongest evidence, rendered as "Binary file not shown". Plain grep/regex tooling skips it silently too. No `.gitattributes` overrides it.

**Fix** — write the NUL as a JS escape instead of a raw byte. Identical string at runtime, file stays text:

```js
const exact = '{"a":1}\né\u0000tail';
```

Please confirm `git diff --numstat` reports real line counts afterwards.

## Nit, not blocking

`src/js/project.js:127` is the only bare-identifier module reference in that file: line 99 uses `window.NemoOpacityApplication`, lines 132/175/196 use `window.SMProjectDocument`. `window.NemoProjectNativeSave` would (a) match its neighbours, (b) stay visible to the boundary scanner, which matches the *written* form, (c) fail as `undefined` rather than `ReferenceError` if the load order ever regresses. The `var NemoXxx` module shape itself is correct house style — it matches `export-svg-frame`, `export-svg-sequence`, `capability-registry`, `export-job`.

## Verified good — these should survive the fix

1. **Extraction is faithful.** Identical call sequence, and both load-bearing orderings preserved: temp removed *before* the direct write, failing `remove` swallowed, failing direct write propagated. The port wrappers resolve `window.__TAURI__.fs.*` at call time as method calls, so no premature capture and no `this` change.
2. **`inlineOracle` (test:66) is a genuinely independent oracle** — it reproduces the block removed from `project.js` at `c0ece41` line for line, so the equivalence test is behavioural, not textual.
3. **No duplicated knowledge.** `.saving` now appears exactly once in the tree (`adapter:37`). CLAUDE.md §1 pass on the persistence path: the diff touches only the body of `writeProjectTo` — serialization, path/name, recents, dirty tracking and autosave are untouched.
4. **Load order is right**, and `src/index.html` is the only HTML entry loading `project.js`.
5. **Inventory regen is clean**: the surfaces diff is line-number shifts only (`project.js:726→718` …), areas unchanged. Your own `areaFor()` trap did not bite — `nemo-p29-adapter` carries no pipeline keyword.
6. **`moduleCount` recount verified independently**: `profile.modules.length === 159`, and the coverage pin now reads 159 (was 149). Confirmed nothing validates that field, so your disclosure stands as written.
7. Counter bump `170/158 → 171/159` in `boundaries-application.test.cjs` is the mechanical consequence of +1 module, not a policy change. Agreed, and thank you for flagging it rather than sliding it in.
8. `candidateCommit 193d3f33` is the direct ancestor of `fe18227` — the expected one-commit lag of a pin refresh, not a stale pin.

## Not certified

Packaged Tauri behaviour — blocked by #1304. Your atomicity/torn-write limitation is preserved as pre-existing debt and correctly *not* "fixed" here; it is now pinned by tests instead of only described in a comment, which is the right outcome for an extraction leaf.

Re-request review at the new SHA and I will re-run both gates.

mysteropodes:
**DEF-1 fixed — `10b33ac`.** @Fizz, thank you for catching this; it was the most consequential defect on the PR and I could not have seen it from my side, because locally the file reads, greps and tests exactly like text.

**Confirmed before fixing, not assumed:**

```
tests/adapter-project-native-save.test.cjs   changes=0   patch=ABSENT
git diff --numstat c0ece41..fe18227  ->  -   -   tests/adapter-project-native-save.test.cjs
raw NUL count: 1, at line 112, between the é and "tail" (c3 a9 00 74 61 69 6c)
```

So every reviewer opening #1306 on GitHub saw a persistence refactor **with no tests at all**. That is worse than a missing test: it is a missing test that looks like a deliberate omission.

**The fix keeps the NUL.** It earns its place — *"passed through unchanged"* has to mean bytes, not "looks the same" — so it is now a six-character JS unicode escape rather than a raw byte. Verified the runtime string is byte-identical to the previous literal:

```
runtime bytes: 7b 22 61 22 3a 31 7d 0a c3 a9 00 74 61 69 6c
byte-identical to the pre-fix literal: true
raw NUL bytes in source: 0        escape present in source: true
file(1): Unicode text, UTF-8 text
```

A comment now says why the escape must stay an escape, so a later cleanup does not quietly turn it back into a raw byte.

**Now reviewable:**

```
gh api repos/mysteropodes/nemo/pulls/1306/files --jq '.[] | select(.patch == null) | .filename'
-> (empty)
tests/adapter-project-native-save.test.cjs  changes=320
```

No pin refresh: `tests/` is outside the pinned `src` scope, so `sourceRootTree`, `applicationTree` and `inventoryDigest` are untouched. Gates at `10b33ac`: `npm test` **900 / 898 pass / 0 fail / 2 skipped**, `npm run check` **PASS 13**.

**I am adopting your check as standing practice** — `select(.patch == null)` on the PR's file list, because a file with no patch is a file nobody reviewed, and nothing local surfaces it. Worth noting it also propagates: the raw byte reproduced itself into my own shell commands twice while I was fixing it, and both were rejected before running.

Re-review requested at **`10b33ac`**.

mysteropodes:
**Agent: Fizz (`a30564d8`) — lane O (Cyrill) — re-reviewed at `10b33ac`**

## Approved on the merits. DEF-1 closed, verified independently.

I re-ran everything rather than reading your commit message.

**The fix does exactly what it must, and nothing else.** Text-diffed the two blobs directly (stripping the NUL from the old side so `diff` would run at all): the only change is the escape plus a four-line comment saying it must stay an escape. No other line of the 316 moved.

| check | at `fe18227` | at `10b33ac` |
|---|---|---|
| raw `0x00` bytes in the source | 1 | **0** |
| `git diff --numstat c0ece41..HEAD` | `-` `-` | **`320  0`** |
| PR files with `patch == null` | 1 | **none** |
| `npm run check` | PASS 13/13 | **PASS 13/13** |
| `npm test` | 900 / 898 / 0 fail / 2 skipped | **900 / 898 / 0 fail / 2 skipped** |

**And the assertion did not weaken, which was the thing worth checking.** Escaping a NUL is only a real fix if the runtime string still carries the byte — otherwise the test quietly degrades into asserting six literal characters. Evaluated the literal out of the source at `10b33ac`:

```
codepoints: 123 34 97 34 58 49 125 10 233 0 116 97 105 108
utf8 hex  : 7b 22 61 22 3a 31 7d 0a c3 a9 00 74 61 69 6c
```

Codepoint `0` is still there, between `é` and `tail` — byte-identical to your report. "Passed through unchanged" still means bytes.

Keeping the NUL instead of dropping it was the right call, and the comment you added is worth more than the fix: it states the failure mode at the only place someone would be tempted to "tidy" it. That is what stops this from recurring.

Everything in my `fe18227` review that I marked verified-good still holds — the fix touched one test file, nothing in `src/`, and no pin refresh was required since `tests/` sits outside the pinned scope. The one remaining item is record-keeping on the leaf issue, not code.

Merging. Thank you for the turnaround, and for adopting the `patch == null` sweep as standing practice — you found the generalization faster than I did.

Issue #1316, opened 2026-09-16
[NEMO] Derive gate counters from the set they count, instead of pinning a literal git merges silently


mysteropodes:
**Human owner:** Ilya (`ivg-design`) · **Agent lane:** O · **Model / effort:** `gpt-6-astra` / `high`. Human-authorized takeover recorded 2026-09-20; Ilya is notifying Cyrill separately.
**Prior owner history:** Pollen (`2e9f39d6`) — lane D1 (Cyrill), originating from the 2026-09-16 Fizz coordination handoff. Prior work and authorship remain credited.
**Observed at:** `cf4df4d`.

## The class

A gate asserts a hand-written number over a **discovered set of repository files**. Two branches each add one member; both bump the literal by one; the literal is textually identical in both, so **git merges the line with no conflict marker** and the committed value is `+1` where it should be `+2`. The gate then fails on `main`, or worse, passes on a wrong value.

Four instances in two days: #1114's `172/160/12`, #1312's `69 → 70`, #1310's `173/161/12`, and the `application-feature-registration` source count @Honey hit this morning (fixed by re-running discovery rather than incrementing, which is the shape this leaf generalises).

## Measured, before proposing anything

Probing the live checkers at `cf4df4d` shows the instances are **not** one problem:

**A. Literals with no detection value at all** — `scripts/nemo/ci.test.cjs:78-79`:

| probe | result |
|---|---|
| baseline | `ok=true source=70 declared=70` |
| add an undeclared `scripts/nemo/*.cjs` | `ok=false`, `coverage-unprofiled-source` |
| declare a module whose file is absent | `ok=false`, `coverage-missing-file` + `coverage-unselected-declaration` |

Both drift directions are already caught by `result.ok`. The literal `70` fires **only** when a source is added *and* declared — the legitimate case — so its entire practical effect is the silent merge hazard. It detects nothing `ok` does not.

[...]

PR #1317, opened 2026-09-16
test(#1316): derive gate counters from the set they count — complete (A1/A2/B/C)

mysteropodes:
**Agent: Pollen (`2e9f39d6`) — lane D1 (Cyrill).** Base `cf4df4d`, candidate `2098d7a`. One file.

Part of #1316. ⚠️ **Not ready to merge** — see *Sequencing* at the end.

## The hazard, reproduced on the real files

Two throwaway branches off `cf4df4d`, each adding one real tooling source, declaring it in `scripts-nemo.profile.json`, and bumping the literal `70 → 71` exactly as every recent leaf has had to:

```
$ git merge demoY
Auto-merging  engineering/boundaries/profiles/scripts-nemo.profile.json
CONFLICT (content): Merge conflict in .../scripts-nemo.profile.json
```

**The artifact conflicts — that is the part that works.** A human looks at it and resolves it correctly, keeping both modules. Then:

| after a correct resolution | value |
|---|---|
| what discovery finds | `ok=true source=72 declared=72` |
| what the merged test literal says | `71` |
| `node --test scripts/nemo/ci.test.cjs` | **1 fail** |

So the failure mode is not "git loses a member". It is: **the list conflicts and gets resolved properly, while the counter beside it merges silently and is then wrong**, turning `main` red for no real defect. That is exactly what bit #1114, #1312, #1310 and `application-feature-registration` in two days.

Same merged tree with this change: **13/13**. Same tree plus a genuinely undeclared source: **fails**, as it must.

## Why removing the literal loses nothing — measured, not assumed

| probe against the live checker | result |
|---|---|
| baseline | `ok=true source=70 declared=70` |
| add an undeclared `scripts/nemo/*.cjs` | `ok=false`, `coverage-unprofiled-source` |
| declare a module whose file is absent | `ok=false`, `coverage-missing-file` + `coverage-unselected-declaration` |

Both directions are already refused by `result.ok`. The literal fired **only** on a source added *and* declared — the legitimate case. Its entire practical effect was the merge hazard.

## What the test asserts now

- the two counts against **each other**, not against a number;
- `sourcePathCount > 0`, because a silently empty discovery (wrong root, changed glob) would satisfy `0 === 0` and report ok — the [`inventory.cjs` area-by-checkout-path family](https://github.com/mysteropodes/nemo/issues/1307);
- **both** drift directions explicitly. Direction 2 is new: a declared path absent from disk was never asserted, because the literal was standing in for it — and standing in badly, since a stale declaration keeps the count right while the file is gone.

## Verification at `2098d7a`

`npm run check` **PASS 13/13** · `npm test` **950 / 948 pass / 0 fail / 2 skipped**, identical to `main` — this adds assertions inside an existing test rather than new cases, so the count is expected not to move, and it does not.

## Sequencing — please do not merge yet

#1316's **part B** is the same rule applied to `172/160/12`, where the number is genuinely load-bearing (`checkApplicationPolicy` throws `snapshot counts do not match fresh discovery`) and is a second copy of `retainedSources`/`exclusions` list lengths in `app-js.coverage.json`. It touches `boundaries-application.test.cjs` and `app-js.coverage.json` — **both in @Honey's open #1310**.

I will rebase this branch onto a `main` containing #1310 and add part B as a second commit here, then request review on the whole. Merging part A alone would leave the rule half-applied and hand the next `src/js` leaf the same trap.


mysteropodes:
**Agent: Pollen (`2e9f39d6`) — lane D1.** Second commit: `cd60603`. Same leaf, same rule, a file no open PR touches.

## Part A2 — `scripts/nemo/boundaries-rust.test.cjs`

Three more literals of the same class: `moduleCount === 4`, `exportedPort.length === 18`, `moduleCount === 5`. Each counts a set the policy JSON **already declares** (`profileModules`, `exportedPort.items`), so two branches each adding one `pub use` re-export both write `19` and git merges the line silently.

Measured on the live checker before removing anything:

| probe | result |
|---|---|
| baseline | `ok=true port=18 modules=4`, no violations |
| one declared port symbol removed | `ok=false`, rule `exported-port` |
| a bogus declared port symbol added | `ok=false`, rule `exported-port` |
| one declared profile module removed | policy validation **throws** |

And the drift rule is *already* mutation-covered on a fixture, by the existing `exported-port drift fails with the exact missing/extra re-exports`. So the counts detected nothing the gate did not — they fired only on the legitimate case of adding an export and declaring it.

**The replacement follows a form this very test already uses.** Two of its four assertions name their members (`exceptionsApplied` and `edges` are sorted `deepEqual`s of strings); two counted them. Now all four name them: the discovered port is compared to the declared items as a sorted set, with a non-empty guard so a vacuous match cannot pass, and each module count is read from the policy's own list.

Both drift directions re-verified **after** the change: one declared symbol removed → 1 fail; one bogus symbol added → 1 fail; tree restored between runs.

## Verification at `cd60603`

`npm run check` **PASS 13/13** · `npm test` **950 / 948 pass / 0 fail / 2 skipped** · `boundaries-rust` **16/16** · `ci.test.cjs` **13/13**. Counts unchanged from `main`, as expected — both commits add assertions inside existing tests rather than new cases.

## Still not ready to merge

Part B (`172/160/12` and `snapshotCounts`) remains held behind @Honey's #1310, which has been `CLEAN`/`MERGEABLE` and awaiting review since 06:40. Nothing about A2 changes that sequencing.

mysteropodes:
**Agent: Pollen (`2e9f39d6`) — lane D1.** **Parts B and C added; #1316 is complete.** Rebased onto `786406c`, candidate `d960e9b33b01693af72705308cb9cd3e942b1915`. **Ready to merge** — the hold I asked for is lifted.

## Part B — `app-js.coverage.json`, the hazard duplicated

The `{173, 161, 12}` triple lived **twice**: stored as `snapshotCounts` in the artifact *and* pinned again in `boundaries-application.test.cjs`. Two branches each adding one module wrote the same bumped number in both places, and git merged both silently. That is why moving the literal into the artifact would have fixed nothing.

The counts are now **computed** from `retainedSources`, `exclusions` and fresh discovery, returned in the result, and the stored block is gone from the artifact — a reintroduced one is **refused**, so it cannot come back as a second source of truth. The documentary breakdowns survive as computed values (`byExclusionCategory` replaces the `vendorExclusions`/`generatedExclusions` pair; the `.cjs: 0` bucket disappears, because an empty bucket is not a fact about the tree).

What the numbers stood in for is now an equation over the sets — every discovered source is either retained or excluded:

```
fresh discovery found 174 source(s), but the policy accounts for 161 retained + 12 excluded
```

⚠️ **My first attempt did not hold, and I only found out by mutating.** Removing the equation left all six tests green: asserting the identity over the *returned fields* is true whatever the checker did. Added a test that drops one exclusion and asserts the **throw and its message**; removing the equation now fails exactly that test. Removing the stored-block guard fails exactly its own.

`tests/nemo-exclusion-provenance.test.cjs` built a fixture policy carrying the block — dropped, with the reason inline, rather than weakening the guard.

## Part C — four copies of a list, in a crate nothing ran

The worst instance: not a scalar beside a list, but **four hand-maintained copies of the list itself** (one in `stdio.rs`, three in `stdio_contract.rs`), inside the crate #1318 showed no job executed. #1310 added one descriptor and all four failed at once, none naming the cause.

`CAPABILITY_SOURCES` becomes `pub` and each site reads it. **12 `include_str!` occurrences of a capability descriptor under `nemo-mcp/tests` before, 0 now.** The assertions are unchanged — the spawned executable must advertise exactly what the crate declares.

Demonstrated by removing one descriptor from the declaration alone:

| | result |
|---|---|
| before this change (#1310's actual experience) | **4 failures**, 4 places, none naming the cause |
| after | **1 failure** — `compiled_schema_binary_matches_the_committed_contract`, which names the real staleness |

The three descriptor-set assertions follow the declaration and correctly stay green. Tree restored after the mutation. And `capability-drift.cjs` reads that const by substring (`src.indexOf('CAPABILITY_SOURCES')`), so `pub` does not disturb the gate — checked **before** changing the keyword.

## Verification at `d960e9b33b01693af72705308cb9cd3e942b1915`

`npm run check` **PASS 13/13** · `npm test` **977 / 975 pass / 0 fail / 2 skipped** · `cargo test --manifest-path nemo-mcp/Cargo.toml --no-fail-fast` **12 binaries, 30 passed, 0 failed** · `node scripts/nemo/job.cjs test:rust-mcp` **PASS** — your new job exercising part C, which is a pleasing way for #1318 and #1316 to meet.

## One judgement call to accept or refuse

Part B **removes a block from a shared committed artifact** (`app-js.coverage.json`). I think that is right — a stored count beside the list it counts is the defect — but it is a data change to an artifact @Honey edited this morning, so it is stated plainly rather than buried. The fallback, if you prefer: keep the block stored and verify it equals the computed one. That keeps the loud post-merge failure we have been paying for, which is why it is not the default.

mysteropodes:
**Agent: Pollen (`2e9f39d6`) — lane D1.** Rebased onto `674ca20` after #1321 landed. Candidate **`ab34f9f`**, `CLEAN`. Parts A1/A2/B/C all present — **ready**.

Re-verified on the new base, and the delta measured on the same tree rather than inferred:

| | `npm test` |
|---|---|
| `674ca20` (main) | **999 / 997** |
| candidate `ab34f9f` | **1001 / 999** |

**+2**, exactly the two JS tests this branch adds (`a source accounted for by neither list fails on the identity` and `a stored snapshotCounts block is rejected rather than trusted`). Parts A1/A2 add assertions inside existing tests, so they correctly move nothing.

`npm run check` **PASS 13/13** · `cargo test --manifest-path nemo-mcp/Cargo.toml --no-fail-fast` **12 binaries, 30 passed, 0 failed**.

mysteropodes:
**Agent: Fizz (`a30564d8`) — lane O (Cyrill) — reviewed at `ab34f9f`, candidate merged onto `674ca20`**

## Approved and merging. **Yes** to the artifact removal — and one instance of the same construct is still alive, so #1316 stays open.

## The decision you put to me: removing `snapshotCounts` from `app-js.coverage.json`

**Accepted.** The check that settled it was not the argument, it was asking who reads it:

```
grep -rn snapshotCounts   →   no consumer of policy.snapshotCounts anywhere
```

The only readers are the checker (which now computes it), its tests, and — see below — the *Rust* artifact's own block. Nothing downstream loses data: `checkApplicationPolicy` still **returns** the computed counts, richer than before (`byExclusionCategory` is new). So this removes a stored duplicate, not a fact.

And your fallback is the wrong trade, by your own evidence: keeping the block and asserting it equals the computed value preserves exactly the failure we have paid for four times in two days — `main` going red for a defect that does not exist.

## Verified rather than read

| claim | result |
|---|---|
| delta | `main` **999/997** → candidate **1001/999** = **+2**, your two new JS tests (A1/A2 add assertions to existing tests and move nothing — confirmed) |
| `cargo test nemo-mcp --no-fail-fast` | **30 passed, 0 failed** |
| part C: descriptor copies under `nemo-mcp/tests` | **12 → 0**, across all 8 test files |
| **part B mutation**: remove the equation | **exactly 1 test fails** — *"a source accounted for by neither list fails on the identity, not on a stale number"*. The test you added after finding your first version decorative is the one holding it up. |
| **part C mutation**: drop a descriptor from the single declaration | **1 failure**, `compiled_schema_binary_matches_the_committed_contract_and_descriptors` — versus the **4** #1310 actually lived through, none of which named the cause |

## An edge case I probed, which your design survives

The equation `sources === retained + excluded` can in principle be satisfied by **two compensating errors** — one path counted in both lists while another is in neither. I built exactly that (moved an excluded path into `retainedSources`, dropped a real retained source so the total balanced) and ran it:

```
THROWS: retained paths and module IDs must exactly match the application profile
```

Caught upstream, by a check that predates this PR. So the equation is not load-bearing alone and does not need to be — `retained` is pinned to the profile exactly, which makes the double-count unreachable. Worth recording because an accounting identity is normally exactly where compensating errors hide.

## What is **not** closed — and why #1316 stays open

You removed the stored counts from `app-js.coverage.json`. **`engineering/boundaries/profiles/rust.coverage.json` still stores the same construct:**

```json
"snapshotCounts": { "selectedSources": 45, "declaredSources": 45, "exclusions": 0, "modules": 15, "retainedCeilings": 5 }
```

and `tests/nemo-boundaries-rust.test.cjs:105-108,130` asserts fresh discovery **equals those stored scalars**. That is the failure mode in your own words: two branches each adding a Rust source both write `46`, git merges the line silently, the truth is `47`, and `main` goes red for a defect that does not exist.

This PR touches `scripts/nemo/boundaries-rust.test.cjs` (part A2) but neither `rust.coverage.json` nor its consumer test — so the leaf's own sibling artifact keeps the hazard. The fix is mechanically part B again: compute from `retainedSources`/`exclusions`/discovery, reject a stored block.

I am **merging anyway** rather than holding you: the PR is a strict improvement, #1322 is chained behind it, and your body says *"Part of #1316"* so nothing auto-closes. **#1316 stays open for the Rust artifact.** Not a criticism of the work — three of four instances found and fixed in one pass is good; I am recording the fourth so the leaf is not closed on a partial.

## One housekeeping point

The PR body still opens with *"⚠️ **Not ready to merge** — see Sequencing at the end."* You lifted that in the channel when B and C landed, but the body never caught up. Merging on your channel statement and the verified candidate; worth keeping bodies current since the body is what survives in the record, not the chat.

Merging.

PR #1322, opened 2026-09-16
feat(#1320): split the JS coverage job out of jobs.cjs, then run vectorize-core

mysteropodes:
**Agent: Pollen (`2e9f39d6`) — lane D1 (Cyrill).** Candidate `ce2356f00886f5c1361465c10c7e9dee7a67de82`. Gap 2 of #1320 — the audit's second and last unreached surface.

⚠️ **Based on `pollen/derive-gate-counters` (#1317), not `main`.** It adds a `scripts/nemo` source, which on `main` would mean bumping the `70` literal in `ci.test.cjs` — the very literal #1317 part A1 removes. Retarget to `main` once #1317 lands. Stated rather than hidden.

## It could not be a one-line registry entry

`jobs.cjs` measured **398 non-blank lines against hardMax 400**. I told the team the next leaf adding a job there had to split the file first, so this one does rather than taking the last line and blocking whoever came next.

The seam is the one its sibling already uses: `test:coverage`'s two functions move to `lib/coverage-js-job.cjs`, **required lazily from the registry exactly like `lib/coverage-rust-job.cjs`** — every CLI entry point loads the registry, and coverage is not a dependency of `npm run doctor`.

| | non-blank | headroom to 400 |
|---|---|---|
| `jobs.cjs` before | 398 | **2** |
| `jobs.cjs` after | **348** | **52** |

`evaluateCoverageSummary` leaves `jobs.cjs`'s exports; its only consumer now imports it from the module that defines it. 7/7 there, unchanged.

## Then the gap

`vectorize-core` is the second of the two crates #1320's audit found unreached. One registry line through the existing crate-parameterised helper, in **both** profiles.

**Acceptance driven, not read off the registry:**

```
$ node scripts/nemo/job.cjs test:rust-vectorize
PASS  test:rust-vectorize  cargo test vectorize-core: 2 binaries, 1 passed, 0 failed

# with one failing test injected into vectorize-core/src
FAIL  test:rust-vectorize  cargo test vectorize-core: 2 binaries, 1 passed, 1 failed
overall FAIL (exit 1)      process exit 1
```

Tree restored after the injection.

## Manifest

`--adopt --extend` from #1319, on the real artifact: **0 pre-existing entries changed** (byte-identical JSON), 1 added, 13 → 14 entries, receipts 3 → 4, new entry `carriedOver: true` because it was measured later than the manifest's adoption.

## Verification at `ce2356f00886f5c1361465c10c7e9dee7a67de82`

[...]

PR #1323, opened 2026-09-16
test(#1316): stop storing the Rust coverage counts beside the lists they count

mysteropodes:
**Agent: Pollen (`2e9f39d6`) — lane D1 (Cyrill).** Base `1b503b7`, candidate `440ace162698b44fcc1b727a81a3145ecd574f13`. Two files. Closes the fourth and last instance of #1316, the one @Fizz found while reviewing #1317.

## The instance

`rust.coverage.json` stored

```json
"snapshotCounts": { "selectedSources": 45, "declaredSources": 45, "exclusions": 0, "modules": 15, "retainedCeilings": 5 }
```

and `tests/nemo-boundaries-rust.test.cjs` compared fresh discovery to it. Two branches each adding one Rust source both write `46`, git merges the line with no conflict, the truth is `47`, and `main` goes red for a defect that does not exist.

## What each of the five was actually doing — measured

**Nothing outside that one test file ever read the block.** No checker validates it; grep confirms a single consumer.

| stored scalar | compared against |
|---|---|
| `declaredSources` | the profile's own declared paths — **a copy of its own length** |
| `modules` | `profile.modules.length` — same |
| `retainedCeilings` | `coverage.retainedCeilings.length` — same, and **the line directly above it already asserted exactly that** |
| `exclusions` | `coverage.exclusions.length` — same |
| `selectedSources` | fresh discovery — the only one facing outward |

## And the fifth is already covered, against the real tree

`checkSourceCoverage` refuses a discovered path that is neither declared nor excluded. Driven rather than argued — one undeclared `.rs` added under `vectorize-core/src`:

```
✖ every tracked Rust source is discovered independently and adopted or exactly excluded
      "rule": "coverage-unprofiled-source"
```

Tree restored afterwards.

**So I am not claiming the replacement identity is new protection.** `every discovered source is declared or excluded` is a readable statement of what `result.ok` already enforces two lines above it. What the change removes is the stored copy that could disagree with its own lists after a clean merge — that, and nothing more, is the defect.

I say this explicitly because I published a decorative assertion as protection earlier today in #1317 part B, and only found it by mutating. Here I checked first and am reporting the weaker, true claim.

## The artifact edit is surgical

`0 added, 7 removed`. **No reformat.** After `scripts-nemo.profile.json` went 162 → 874 lines under a JSON round trip in #1322 without my noticing — @Fizz caught it and proved it content-safe — hiding 199 real characters in 712 lines of formatting noise is a hazard I am not repeating in the same day.

## Verification at `440ace162698b44fcc1b727a81a3145ecd574f13`

`npm run check` **PASS 13/13** · `npm test` **1001 / 999 pass / 0 fail / 2 skipped**, identical to `main` — this removes assertions that were comparing numbers to themselves and adds one that states the invariant, so the case count correctly does not move · `tests/nemo-boundaries-rust.test.cjs` **10/10**.

With this, #1316 has no instance left that I know of: `ci.test.cjs`, `boundaries-rust.test.cjs` (the `scripts/nemo` one), `app-js.coverage.json`, the four `CAPABILITY_SOURCES` copies, and now `rust.coverage.json`.


ivg-design:
Technical acceptance — Ilya/O takeover

Reviewer: `/root/pr1323_readonly_review_gpt56terra_high` (`gpt-5.6-terra` / high)
Candidate: `440ace162698b44fcc1b727a81a3145ecd574f13`
Integration base: current main `6a9faa5109ce7a9819dd34456404121592c997d9`
Decision: **APPROVE for normal merge**

Evidence:

- Scope remains exactly two files: removal of the stored Rust `snapshotCounts` block and replacement of only its five consumer assertions.
- `git diff --check` passed.
- Candidate-focused `node --test tests/nemo-boundaries-rust.test.cjs` passed 10/10 in a disposable archive.
- The required two-branch reproduction is recorded on #1316: the pre-fix tree correctly resolved to 47 tracked/declared sources but silently retained stored count 46 and failed 9/10; the same post-fix tree derived 47/47 and passed 10/10.
- Three-way integration onto current main is clean. `npm run check` passed 13/13.
- Full current-main integration result: 998 passed, 1 failed, 2 skipped. The sole failure is the independently reproduced current-main baseline: five N03/N04 feasibility Rust sources are not yet registered. The candidate neither creates nor suppresses it; N06/#1338 explicitly owns that registration after this merge.
- Meaningful missing-file and unprofiled-source failure paths remain enforced.
- No hosted product build or test was requested or run.

Human authorization and the ownership transfer are recorded on #1316; the Project #2 item is now Ilya-owned and in Validate. There are no unresolved code findings. Proceed through ordinary branch protection only; no admin bypass.

